PR #4, opened 2021-01-22
Add etl-graph as a job via submodule

acmiyaguchi:
The [etl-graph](https://github.com/mozilla/etl-graph) repo powers https://etl-graph.protosaur.dev. I thought it might be interesting to add it to docker-etl so I don't have to set up circle-ci with GCR and the appropriate contexts. I currently have an open PR to add docker support (https://github.com/mozilla/etl-graph/pull/8), and all I would need to do is add the two yaml fragments for scheduling here.

It does mean that that updating the container is a two step process of PR in the submodule, and then updating this repo with the new commit to build the container. [Dependabot has submodule support](https://dependabot.com/submodules/), so this could be less onerous to keep up to date. 

@BenWu do you have particular thoughts about whether this workflow is appropriate in this context. I have another job that I think would fit this submodule pattern (https://github.com/mozilla/mozaggregator2bq) and would save a bit of time with the circleci config.

### DIFF
--- a/.circleci/config.template.yml
+++ b/.circleci/config.template.yml
@@ -31,6 +31,9 @@ jobs:
       - image: docker:stable-git
     steps:
       - checkout
+      - run:
+          name: Checkout git submodules
+          command: git submodule update --init --recursive
       - setup_remote_docker:
           version: 19.03.13
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,6 +34,9 @@ jobs:
       - image: docker:stable-git
     steps:
       - checkout
+      - run:
+          name: Checkout git submodules
+          command: git submodule update --init --recursive
       - setup_remote_docker:
           version: 19.03.13
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,6 +68,24 @@ jobs:
           name: Test Code
           command: docker run app:build pytest --flake8 --black
 
+  # As seen from the perpsective of https://github.com/mozilla/docker-etl
+  build-job-etl-graph:
+    docker:
+      - image: docker:stable-git
+    steps:
+      - checkout
+      - run:
+          name: Checkout git submodules
+          command: git submodule update --init --recursive
+      - compare-branch:
+          pattern: ^jobs/etl-graph/
+      - setup_remote_docker:
+          version: 19.03.13
+      - run:
+          name: Build Docker image
+          command: docker build -t app:build jobs/etl-graph/
+
+
   build-job-example_job:
     docker:
       - image: docker:stable-git
@@ -113,6 +131,20 @@ workflows:
           filters:
             branches:
               only: main
+
+  job-etl-graph:
+    jobs:
+      - build-job-etl-graph
+      - gcp-gcr/build-and-push-image:
+          context: data-eng-airflow-gcr
+          path: jobs/etl-graph/
+          image: etl-graph_docker_etl
+          requires:
+            - build-job-etl-graph
+          filters:
+            branches:
+              only: main
+
 
   job-example_job:
     jobs:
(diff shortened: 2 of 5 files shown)


BenWu:
I think this is an appropriate use of this repo.  This is possibly better than keeping the job code here as I originally intended since we get the benefits of separate repos like access management.  As long as dependabot works well I think this can be a pretty good workflow so definitely worth trying.

jklukas:
I do worry that this muddies the contract about what this repo is for. Introducing the concept of git submodules could make this repo more surprising to work on.

Can we move the source code for etl-graph here instead and tear down the separate repo?

acmiyaguchi:
> I do worry that this muddies the contract about what this repo is for. Introducing the concept of git submodules could make this repo more surprising to work on.

Could you go more into depth on how this would be surprising? The Github UI for submodules is pretty clear about showing links.  `docker-etl` would still be a monorepo that indexes source code and builds/publishes images on directory diffs, but without having to munge development history. If it doesn't make sense to have a separate repository, then it can also go directly here. I think the following repositories could fit a submodule pattern.

- https://github.com/mozilla/etl-graph
- https://github.com/mozilla/mozaggregator2bq
- https://github.com/mozilla/webrender_intel_win10_nightly
- https://github.com/mozilla/forecasting (although this is nested an extra layer deeper)

These are small repositories that are dockerized and benefit from being scheduled on Airflow. Copying source here is certainly an option, but it loses history. In the case of etl-graph for example, I think it's clearer to check out the canonical repo and run `npm install && npm run dev` without having to worry about the gigantic directory structure imposed here for the benefit of automation. 

I think this probably goes

jklukas:
> Could you go more into depth on how this would be surprising?

Mosty that the documentation in the README assumes that the job code lives here. I think we would need to update that to make it clear that submodules are an option and what the workflow is like when using them.

Also, the [proposal doc](https://docs.google.com/document/d/1r7D6S32DczDtYcVRKf6RZ-k_FjxP0vktBQsy7I7GSXQ/edit) specifies a permissions model for this repo:

> Due to the elevated permissions granted to tasks running in Airflow, the scheduled jobs will have the ability to perform actions that the user cannot perform, such as reading and writing to restricted GCP resources.  Therefore, all ETL jobs must be properly vetted.  The proposed workflow is that all jobs and changes to those jobs must be reviewed by a data platform engineer

How would this submodule approach affect the permissions model? Can we add a PR checklist to ensure reviewers consider the permissions implications of their review?

acmiyaguchi:
> Mosty that the documentation in the README assumes that the job code lives here. I think we would need to update that to make it clear that submodules are an option and what the workflow is like when using them.

This is something that I can make clear in this PR. There are a couple of things that need to be explored, like the use of dependabot for immediate PRs based on submodule updates.

> How would this submodule approach affect the permissions model? Can we add a PR checklist to ensure reviewers consider the permissions implications of their review?

The permissions model is still in place, although the layer of indirection could inadvertently facilitate rubberstamped r+'s. There are multiple layers of reviews where concerns can be appropriately raised by a reviewer -- first in the source repository, then in the PR to include it in docker-etl, and finally in PR to actually schedule the job in telemetry-airflow. In the case of a collaboration with data science, a data engineer is typically involved when scheduling code on airflow anyhow.

Presumably, r+ing a dependabot update to a submodule should involve reviewing changes since the last update as due diligence. A checklist like we do in mozilla-pipeline-schemas is a good idea since we effectively promote source in this repo to run in our production environment.

BenWu (approved):
Looks good.  I'm assuming you'll put together a small proposal with this as a proof of concept.

Are you going to update the submodule after merging the other pr?  Or see what dependabot does?

acmiyaguchi:
I do have a proposal I'll share around tomorrow at the infrastructure working group covering some of the discussion here.  I had to merge the other PR first so the tests pass.  I swapped branches to main, so I want to make one more PR to make sure that dependabot works as expected.

PR #5, opened 2021-01-27
Add dependabot support for git submodules

acmiyaguchi:
It turns out that dependabot doesn't have live updates for submodules, which puts a hamper on efforts of using submodules. I still want to try this out though, so I've set updates to occur daily 2 hours before most of our airflow jobs start.

Checklist for reviewer:

  responsibly. If this is an update to a submodule, take the time to review the
  explicit diff of the submodule.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "gitsubmodule"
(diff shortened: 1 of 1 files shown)


BenWu (approved):
Yeah daily update seems inconvenient but I agree it's worth trying

PR #10, opened 2021-01-28
Migrate repo example

BenWu:
Just an example.  History is intact

Command I ran was `git subtree add -P jobs/etl-graph ../etl-graph main`

PR #11, opened 2021-01-28
Import etl-graph as subtree instead of git module

acmiyaguchi:
Here's a copy of https://github.com/mozilla/docker-etl/pull/10, with some minor modifications to reset the ci fragments to the default versions without any workspace shenanigans. I'll archive the old repo and redirect any documentation here to make it the canonical source of info.

Checklist for reviewer:

- [x] Scan the PR and verify that no changes (particularly to
- [x] Ensure the container image will be using permissions granted to
  responsibly. If this is an update to a submodule, take the time to review the
  explicit diff of the submodule.


BenWu (approved):
Thanks for going through with this experiment with submodules.  The proposal did a good job of pointing out the positives and the negatives.  This looks good to me but do you think this warrants another (tiny) proposal?  Maybe there's something were missing.

Can you also take out the submodule sentence in the pr template?

BenWu:
Oh one issue here is that the github pr references in the merge commits are pointing to this repo.  That's an important downside.  I wonder if it's possible to rewrite those easily to point to etl-graph

acmiyaguchi:
> This looks good to me but do you think this warrants another (tiny) proposal? Maybe there's something were missing.

It's a good idea, if for nothing else, to gather consensus on importing repositories as subtrees. The experience of trying out the workflow and writing a proposal has been a learning experience for me. I do want to get this container built and scheduled though at this point so I can make incremental updates again, so I'm going to merge this in. I'll probably have another proposal for next week's meeting along with the rejection of submodules.

Alongside with the issue numbers being misleading in the commit histories, I notice that subtrees require merge commits instead of rebase commits due to the way that they are structured. I'm not into the idea of a squash commit for the initial import.

acmiyaguchi:
One behavior of `git subtree add` that didn't work as expected is that the history is appended as if it were still at the root of the repository, and then finally merged into the repository (5819d47). This means that blame won't work for historical purposes outside of the commit log. I'd recommend squash commits or a script that replays commits into a subdirectory (and possibly rewrites commit messages to point to the original repository). The number of repositories that would want to be merged into docker-etl is a small handful though, so the limitations are not worth spending too much energy on.

PR #29, opened 2021-11-16
Road 134 - Introduce a job to mirror suggestions from RS to BigQuery

Dexterp37:
This introduces the job required for [this ticket](https://mozilla-hub.atlassian.net/browse/ROAD-134)

This is a draft because I want to test it locally first and I had troubles with the GCP/BigQuery interaction.

Checklist for reviewer:

  responsibly.


jobs/quicksuggest-2-bq/README.md
```diff
@@ -0,0 +1,45 @@
+# Python Template Job
```
jklukas, line 1: Note to fill in some docs here.

jobs/quicksuggest-2-bq/ci_job.yaml
```diff
@@ -0,0 +1,18 @@
...
+        version: 19.03.13
+    - run:
+        name: Build Docker image
+        command: docker build -t app:build jobs/quicksuggest-2-bq/
+    - run:
+        name: Test Code
+        command: docker run app:build pytest --flake8 --black
+    - run:
+        name: Lint Code
+        command: docker run app:build flake8
```
jklukas, line 18:
```suggestion
        command: docker run app:build flake8

```

jobs/quicksuggest-2-bq/ci_workflow.yaml
```diff
@@ -0,0 +1,13 @@
...
+    - gcp-gcr/build-and-push-image:
+        context: data-eng-airflow-gcr
+        docker-context: jobs/quicksuggest-2-bq/
+        path: jobs/quicksuggest-2-bq/
+        image: quicksuggest-2-bq_docker_etl
+        requires:
+          - build-job-quicksuggest-2-bq
+        filters:
+          branches:
+            only: main
```
jklukas, line 13:
```suggestion
            only: main

```

jobs/quicksuggest-2-bq/quicksuggest_2_bq/main.py
```diff
@@ -0,0 +1,207 @@
...
+            bigquery.SchemaField("iab_category", "STRING"),
+            bigquery.SchemaField("icon", "STRING"),
+            bigquery.SchemaField("impression_url", "STRING"),
+            bigquery.SchemaField("id", "INTEGER"),
+            bigquery.SchemaField("keywords", "STRING", "REPEATED"),
+            bigquery.SchemaField("title", "STRING"),
+            bigquery.SchemaField("url", "STRING"),
+        ],
+        source_format=bigquery.SourceFormat.NEWLINE_DELIMITED_JSON,
+        write_disposition="WRITE_APPEND",
```
jklukas, line 116: Our ETL best practices include making sure that jobs are idempotent, so we generally use WRITE_TRUNCATE since WRITE_APPEND could lead to duplicates if the job gets rerun. To support having a history, we should add `time_partitioning` here to make this a date-partitioned table. And for consistency with existing practice, I'd recommend naming the date field `submission_date`.

jobs/quicksuggest-2-bq/quicksuggest_2_bq/main.py
```diff
@@ -0,0 +1,207 @@
...
+    load_job = client.load_table_from_json(
+        suggestions,
+        destination_table_id,
+        location="US",
+        job_config=job_config,
+    )
+
+    try:
+        load_job.result()
+    except BadRequest:
```
jklukas, line 128: I think we shouldn't catch this exception and instead let the job fail so that in Airflow it gets marked as a failure.

jobs/quicksuggest-2-bq/quicksuggest_2_bq/main.py
jklukas: I do generally think it's good to maintain a history table for ETL jobs like this, but since it's not possible (I assume) to fetch old versions of content from RemoteSettings, it may be fine to have this table simply represent the current state and not be partitioned. In which case we would not need a `--date` input parameter.

jobs/quicksuggest-2-bq/quicksuggest_2_bq/main.py
```diff
@@ -0,0 +1,207 @@
...
+        for e in load_job.errors:
+            logging.error(f"ERROR: {e['message']}")
+
+    stored_table = client.get_table(destination_table_id)
+    logging.info(f"Loaded {stored_table.num_rows} rows.")
+
+
+@click.command()
+@click.option(
+    "--date",
```
jklukas, line 138: It looks like we don't actually use this. If we're going to include a date field in the output, we should make sure it reflects this.

.circleci/config.yml
```diff
@@ -170,6 +170,22 @@ jobs:
           command: docker run app:build make lint
 
 
+  build-job-quicksuggest-2-bq:
```
jklukas, line 173: My personal preference would be to avoid punctuation and call this job `quicksuggest2bq`. We already have two jobs in this repo that follow that convention. It may make some people's eyes twitch, though 👁️

jobs/quicksuggest-2-bq/quicksuggest_2_bq/main.py
Dexterp37: This is currently being used in `store_suggestions` to augment the parsed suggestions with the date. But we can remove this if we no longer want it as you suggested above!

jobs/quicksuggest-2-bq/quicksuggest_2_bq/main.py
Dexterp37: Good point. I think I need to catch the exception anyway, to make sure we're able to print the errors (if we don't do that, the exception text alone is not really helpful!), but we should indeed raise it again to make sure the job fails.

jobs/quicksuggest-2-bq/quicksuggest_2_bq/main.py
Dexterp37: This all makes sense, thanks for the detailed explanation! I was working under the assumption that there's value in storing the history in the table. While we cannot fetch it from RS itself, [it is available](https://github.com/mozilla-services/quicksuggest-rs/tree/main/source-data) (see the [relevant discussion](https://mozilla.slack.com/archives/G01NU8LUXPH/p1636991319275200?thread_ts=1636983055.268100&cid=G01NU8LUXPH)).

I'm ok with keeping this job simple, i.e. dropping the `--date`, and adding it back if we'll ever need it. Does that sound ok, @jklukas ?

.circleci/config.yml
Dexterp37: @jklukas would you rather have the whole package renamed (both the directory and the files) or just this one job name?

Dexterp37:
@jklukas please also note that, for some reason, CI isn't running on this :| Maybe it's not running for PRs on forks?

.circleci/config.yml
jklukas: Sorry for not being clear there. I would want all the directories and files to consistently use `quicksuggest2bq`

jobs/quicksuggest-2-bq/quicksuggest_2_bq/main.py
jklukas: Yes, that sounds great.

jklukas (approved):
:shipit:

jklukas:
I'm looking at CI now

jklukas:
Indeed I had to push to a branch on this repo to get full CI to run.

jklukas:
Looks like there's a missing lint task.

Dexterp37:
> 
> 
> Indeed I had to push to a branch on this repo to get full CI to run.

Woha, that did the trick! @jklukas is the preferred workflow for this repo to create a branch here rather than have a fork-based workflow?

jklukas:
> is the preferred workflow for this repo to create a branch here rather than have a fork-based workflow

I think you're the first to make a PR from a fork on this repo, so there's been no investment in understanding that workflow. So in practice, yes, working from a branch on this repo is the preferred workflow.

PR #58, opened 2022-06-02
[RS-264] Kpi forecasting non cumulative

perrymcmanis144:
New sql query to do non-cumulative desktop forecasting as per: https://mozilla-hub.atlassian.net/browse/RS-264

Checklist for reviewer:

  responsibly.


jobs/kpi-forecasting/kpi-forecasting/Utils/DBWriter.py
```diff
@@ -110,8 +110,9 @@ def write_confidence_intervals_to_bigquery(
 
     write_table = config["confidences_table"]
     today = str(date.today())
+    metric = f'{config["target"]}_{config["forecast_variable"]}'
 
-    confidences["target"] = config["target"]
+    confidences["target"] = metric
```
scholtzan, line 115: Every run overrides the result table, right? So changing the target name format should be fine.

scholtzan (approved):

jobs/kpi-forecasting/kpi-forecasting/Utils/DBWriter.py
perrymcmanis144: I believe it actually appends; I discussed with @skahmann3 and the consumers of this data are OK with the string values being a bit different going forward, as opposed to making a new column, which would requiring deletion & re-backfilling the old one. 

If this is non-kosher from a DE perspective LMK, but then I will need an assist on the backfilling part.

PR #59, opened 2022-06-15
CONSVC 1818: Include Offline Expansion data for quicksuggest2bq

ncloudioj:
Checklist for reviewer:

  responsibly.

We added a new record type (`offline-expansion-data`) on Remote Settings to support the Firefox Suggest Offline Expansion rollout. Per [SUGGEST-291](https://mozilla-hub.atlassian.net/browse/SUGGEST-291), we'd like to expand this feature to the Merino-backed Firefox Suggestion Online population, which means we need to include keywords from those new suggestions to the allowlist for search term sanitization.     

This fixes [CONSVC-1818](https://mozilla-hub.atlassian.net/browse/CONSVC-1818).

### DIFF
--- a/jobs/quicksuggest2bq/quicksuggest2bq/main.py
+++ b/jobs/quicksuggest2bq/quicksuggest2bq/main.py
@@ -11,7 +11,7 @@ import requests
 from dataclasses import asdict, dataclass
 from google.api_core.exceptions import BadRequest
 from google.cloud import bigquery
-from typing import Dict, List
+from typing import Iterator, List
 
 
 @dataclass
@@ -33,23 +33,23 @@ class KintoSuggestion:
     url: str
 
 
-def download_suggestions(client: kinto_http.Client) -> Dict[int, KintoSuggestion]:
+def download_suggestions(client: kinto_http.Client) -> Iterator[KintoSuggestion]:
     """Get records, download attachments and return the suggestions."""
 
     # Retrieve the base_url for attachments
     server_info = client.server_info()
     attachments_base_url = server_info["capabilities"]["attachments"]["base_url"]
 
-    # Only consider "data" records, search for the following code in Merino
-    # for record in remote_settings_client.records_of_type("data".to_string())
+    # Load records for both "type: data" and "type: offline-expansion-data".
+    # See details in: https://mozilla-hub.atlassian.net/browse/CONSVC-1818
     data_records = [
-        record for record in client.get_records() if record["type"] == "data"
+        record
+        for record in client.get_records()
+        if record["type"] in ["data", "offline-expansion-data"]
     ]
 
     # Make use of connection pooling because all requests go to the same host
     requests_session = requests.Session()
-
-    suggestions = {}
 
     for record in data_records:
         attachment_url = f"{attachments_base_url}{record['attachment']['location']}"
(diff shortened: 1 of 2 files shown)


jobs/quicksuggest2bq/quicksuggest2bq/main.py
```diff
@@ -33,24 +33,24 @@ class KintoSuggestion:
     url: str
 
 
-def download_suggestions(client: kinto_http.Client) -> Dict[int, KintoSuggestion]:
+def download_suggestions(client: kinto_http.Client) -> Iterator[KintoSuggestion]:
```
ncloudioj, line 36: Note that we haven't figured out what IDs to be used for Wikipedia suggestions for Firefox Suggest Offline Expansion, it's currently [hardcoded](https://github.com/mozilla-services/quicksuggest-rs/blob/main/wikipedia/make_suggestions.py#L134) as 0, so I need to change this from Dict to Iterator.

jobs/quicksuggest2bq/tests/test_main.py
```diff
@@ -57,7 +64,7 @@ def test_suggestion_properties_are_properly_parsed(self):
         KintoSuggestion(**SAMPLE_SUGGESTION)
 
     def test_suggestion_download(self, mocked_kinto_client):
-        suggestions = download_suggestions(mocked_kinto_client)
-        assert len(suggestions) == 1
-        assert SAMPLE_SUGGESTION["id"] in suggestions
-        assert suggestions[SAMPLE_SUGGESTION["id"]].title == SAMPLE_SUGGESTION["title"]
+        suggestions = list(download_suggestions(mocked_kinto_client))
+        assert len(suggestions) == 2
```
Dexterp37, line 68: Why is this 2 and not 3?

jobs/quicksuggest2bq/tests/test_main.py
```diff
@@ -57,7 +64,7 @@ def test_suggestion_properties_are_properly_parsed(self):
...
+        assert suggestions[0] == KintoSuggestion(**SAMPLE_SUGGESTION)
+        assert suggestions[1] == KintoSuggestion(**SAMPLE_SUGGESTION)
```
Dexterp37, line 70: Can we change the test a bit so that we get a slightly different `SAMPLE_SUGGESTION` here? Just to make sure we're returning different records.

jobs/quicksuggest2bq/tests/test_main.py
ncloudioj: I added an icon record to `mock_records`, wanted to make sure it won't be downloaded here. I've split it out into its own test to avoid confusion.

jobs/quicksuggest2bq/tests/test_main.py
ncloudioj: Good point. 👍

jobs/quicksuggest2bq/quicksuggest2bq/main.py
```diff
@@ -23,34 +23,34 @@ class KintoSuggestion:
...
     advertiser: str
-    click_url: str
     iab_category: str
     icon: str
-    impression_url: str
     id: int
     keywords: List[str]
     title: str
     url: str
+    click_url: Optional[str] = None
```
ncloudioj, line 32: Our in-house Wikipedia suggestions don't have `click_url` and `impression_url` unlike the adM-provided counterparts.

Dexterp37: Might be worth adding a comment about it above this line, but I'll leave it to your judgement.

jobs/quicksuggest2bq/tests/test_main.py
Dexterp37: Thanks for clarifying!

Dexterp37 (approved):
r+ with an optional comment. Let's also make sure @jklukas is fine with this :)

jklukas (approved):
LGTM

ncloudioj:
Thank you all for the review!

@jklukas, is there anything else we need to update for the search term sanitization job?

jklukas:
For reference, this is how the RS data gets used for sanitization: https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/search_terms_derived/suggest_impression_sanitized_v2_external/query.sql#L65-L85

PR #66, opened 2022-07-27
[WIP] Adds a new job for indexing wikipedia exports into an elasticsearch instance.

quiiver:
Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/wikipedia-indexer/wikipedia_indexer/wiki/settings.py
+++ b/jobs/wikipedia-indexer/wikipedia_indexer/wiki/settings.py
@@ -1,0 +1,149 @@
+SUGGEST_MAPPING = {
+  "dynamic": False,
+  "properties": {
+    "batch_id": {
+      "type": "long"
+    },
+    "doc_id": {
+      "type": "keyword"
+    },
+    "title": {
+        "type": "keyword"
+    },
+    "suggest": {
+      "type": "completion",
+      "analyzer": "plain",
+      "search_analyzer": "plain_search",
+      "preserve_separators": True,
+      "preserve_position_increments": True,
+      "max_input_length": 255
+    },
+    "suggest-stop": {
+      "type": "completion",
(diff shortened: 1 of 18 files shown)


jobs/wikipedia-indexer/wikipedia_indexer/wiki/settings.py
```diff
@@ -0,0 +1,149 @@
...
+      "max_input_length": 255
+    }
+  }
+}
+
+
+SUGGEST_SETTINGS = {
+  "number_of_replicas": "2",
+  "refresh_interval": "-1",
+  "number_of_shards": "8",
```
mtbottle, line 36: I'm curious to how you picked 8 shards for the suggest index. The data, from what I understand, shouldn't be very large, and having more shards generally means more overhead when it comes to searching (as you have to aggregate more things). [They generally recommend that the shards should be about 10-50Gb in size](https://www.elastic.co/guide/en/elasticsearch/reference/current/size-your-shards.html#shard-size-recommendation) to avoid search overhead but doesn't get in the way of efficient re-balancing.

hackebrot (commented):
Hi @quiiver! 👋🏻 

I'd love to chat about automated checks and tests for this ETL job at some point!

Is [DENG-413](https://mozilla-hub.atlassian.net/browse/DENG-413) the corresponding Jira ticket?

jobs/wikipedia-indexer/wikipedia_indexer/wiki/settings.py
quiiver: Originally i was including the intro text in the index and it was in the 40gb range, additionally the cluster size was 4 nodes so I made it a multiple of the cluster size so partitions would be spread across the cluster equally. This can certainly be adjusted

quiiver:
This is now in the merino repository

PR #73, opened 2022-08-15
Introduce job to be run on docker-etl to record data validation metrics

chelseatroy:
# Search Volume Data Validation

This job contains scripts for evaluating whether our recorded search terms 
(candidate search volume for being sanitized and stored) are changing in ways
that might invalidate assumptions on which we've built our sanitization model.

### For Reviewers: 

This is my first time making a PR to this repo, and I'd love help with the following:

1. How can I make sure that this script is working properly? It requires authentication to bigquery, so I'm unclear how to run it locally or check that it will run properly when deployed.
~2. When, exactly, should I expect my tests to run?~ Got this figured out!
~3. How will I go about scheduling this job?~ Got this figured out!

I am happy to pair on any or all of the above. 

Checklist for reviewer:

  responsibly.


jobs/search-term-data-validation/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM python:3.7
```
quiiver, line 1: This is a fairly old version of Python. the Nightly job is targeting python:3 with no minor version specified. That means it targets python 3.10 (according to https://hub.docker.com/_/python).

I would suggest using a newer version of python

jobs/search-term-data-validation/ci_job.yaml
```diff
@@ -0,0 +1,15 @@
+build-job-kpi-forecasting:
```
quiiver, line 1: This name probably needs to be updated.

quiiver: the name here should match whats on this line https://github.com/mozilla/docker-etl/pull/73/files#diff-99009dbadafe387d6577d23b8496536d46d244c3a91bb56943a44d40a32305faR10

jobs/search-term-data-validation/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+services:
+  app:
+    build:
+      context: .
+    volumes:
+      - ./:/app
+      # Mount the local gcloud sdk configuration when developing. Note that this
+      # will modify the contents on the host.
+      - ${CLOUDSDK_CONFIG}/:/tmp/.config/gcloud
+    environment:s
```
quiiver, line 12: is the `:s` correct syntax? it seems to be breaking the yaml parsing in gh but i'm no yaml expert so it could be 🤷

chelseatroy: I more or less copied the structure of `kpi-forecasting`, which has it!

jobs/search-term-data-validation/ci_job.yaml
chelseatroy: Ope, thank you! Good catch.

jobs/search-term-data-validation/src/data_validation.py
```diff
@@ -0,0 +1,421 @@
...
+
+    x_day_moving_average = f"{moving_average_window}_day_{metric}_moving_avg"
+    validation_data[x_day_moving_average] = (
+        validation_data[metric]
+        .rolling(window=moving_average_window, min_periods=0)
+        .mean()
+    )
+
+    comparison_values = validation_data["finished_at"].apply(
+        lambda m: comparison_earliest_date < m.date() <= test_earliest_date
```
perrymcmanis144, line 295: I'm not sure what your dataset size is but maybe worth considering np.where? Definitely not a big enough deal to prevent approval from me though

perrymcmanis144 (approved):
LGTM 👍

PR #81, opened 2022-09-12
Adjust Dockerfile to end up in the right place while installing imports and then running data val job

chelseatroy:
Why we're CDing twice:

1. We need to be in app/ to install the dependencies
2. Then we need to be in app/src/ to run the script itself

### DIFF
--- a/jobs/search-term-data-validation/Dockerfile
+++ b/jobs/search-term-data-validation/Dockerfile
@@ -2,9 +2,12 @@ FROM python:3
 MAINTAINER Chelsea Troy <ctroy@mozilla.com>
 
 WORKDIR /usr/app/src
+WORKDIR /usr/app/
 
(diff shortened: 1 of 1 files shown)


jobs/search-term-data-validation/Dockerfile
```diff
@@ -2,9 +2,12 @@ FROM python:3
 MAINTAINER Chelsea Troy <ctroy@mozilla.com>
 
 WORKDIR /usr/app/src
```
quiiver, line 4: did you mean to replace this line?

chelseatroy: Tried to do this automated merge thing that did not work XD but currently taking a different approach anyway!

perrymcmanis144 (approved):
👍 lgtm, I think this strategy is the correct one at this point

PR #88, opened 2022-10-28
Retrieve job metadata metrics in ascending order to support calculating moving average windows

chelseatroy:
These need to be retrieved in _ascending_ order so that moving average windows are calculated on the five days in the _past_, not the ones in the _future!_ 

Checklist for reviewer:

  responsibly.


perrymcmanis144 (approved):
Does what it says it should, so lgtm 👍

PR #91, opened 2022-11-14
Change ownership of the quicksuggest2bq job

Dexterp37:
I'm no longer involved in this work, so this might benefit from an owner who is more involved with it!

Checklist for reviewer:

  responsibly.


quiiver (commented):
MAINTAINER is deprecated in dockerfiles so we might need a different way to show "ownership" is the parent dag in airflow enough?

PR #92, opened 2022-11-15
Bug #1799081 Add support for full keywords to quicksuggest2bq

quiiver:
Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/quicksuggest2bq/tests/test_main.py
+++ b/jobs/quicksuggest2bq/tests/test_main.py
@@ -19,6 +19,9 @@ SAMPLE_SUGGESTION = {
         "sample da",
         "sample dat",
     ],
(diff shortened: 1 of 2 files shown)


quiiver:
Probably not the most pythonic way to do this but i think it should work?

jobs/quicksuggest2bq/tests/test_main.py
```diff
@@ -19,6 +19,7 @@
         "sample da",
         "sample dat",
     ],
+    "full_keywords": [("sample data", 3)],
```
Dexterp37, line 22: Would you kindly add one more full keyword so that we can test the case of "multiple full keywords" ?

Dexterp37 (approved):

PR #97, opened 2023-01-27
[RS-587] Macroeconomic indices job

curtismorales:
Checklist for reviewer:

  responsibly.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,7 +81,6 @@ jobs:
           name: Test Code
           command: docker run app:build pytest --flake8 --black
 
-
   build-job-example_job:
     docker:
       - image: docker:stable-git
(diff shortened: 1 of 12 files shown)


akkomar (changes requested):
Tests are failing.

Comments:
1. Do we need `pytest` configuration here?
2. Are we sure about `yfinance` licensing? (cc @xluo-ds)
3. Another option for backfilling I though of: in the job check if the table is empty/doesn't exist, if so load historical data.

xluo-ds:
@akkomar  I am not sure about the licensing of `yfinance`, what's the requested proper licensing for our work?

akkomar:
@xluo-ds I think we just need to be sure it's OK to use this for non-personal, commercial purpose.

PR #105, opened 2023-03-22
Add --date parameter to select day and increase timeout

simon-friedberger:
@akkomar As just discussed. This also needs a patch to telemetry-airflow


akkomar (approved):
LGTM, as discussed we should eventually make this idempotent.

PR #115, opened 2023-05-11
make forecast deterministic

bochocki:
Following [this comment](https://github.com/facebook/prophet/issues/1124#issuecomment-812904897), this PR sets a `np.random.seed` to ensure that Prophet forecasts are deterministic (and therefore repeatable). I checked that the forecasts are indeed repeatable by running locally after adding the following line to `kpi_forecasting.py` before results were written to bigquery:

```python
print(predictions.iloc[:, 1:].sum().sum() + confidences.iloc[:, 4:].sum().sum())
```

This sums all of the numeric columns in the `predictions` and `confidences` dataframes to provide a quick check that dataframes are equal across runs. This is not the most exhaustive check that could exist, but imo it's a sufficient demonstration for our use case.

```
# without setting np.random.seed
> 750505443639.0608
> 750445662883.6375
> 750454470768.1648

# after setting np.random.seed
> 750462584055.1995
> 750462584055.1995
> 750462584055.1995
```

Additional Changes:
- Reorder imports

Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/kpi-forecasting/kpi-forecasting/Utils/FitForecast.py
+++ b/jobs/kpi-forecasting/kpi-forecasting/Utils/FitForecast.py
@@ -1,16 +1,18 @@
-from datetime import datetime
+import holidays
+import numpy as np
+import pandas as pd
+import prophet
 import typing
 
-import pandas as pd
-
-import prophet
-
-import holidays
+from datetime import datetime
 
 
 def run_forecast(
     dataset: pd.DataFrame, config: dict
 ) -> typing.Tuple[pd.DataFrame, pd.DataFrame]:
+    # Set random seed to ensure forecast is deterministic
+    np.random.seed(42)
+
     target = config["target"]
 
     fit_parameters = config[
--- a/jobs/kpi-forecasting/docker-compose.yml
+++ b/jobs/kpi-forecasting/docker-compose.yml
@@ -9,6 +9,6 @@ services:
       # Mount the local gcloud sdk configuration when developing. Note that this
       # will modify the contents on the host.
       - ${CLOUDSDK_CONFIG}/:/tmp/.config/gcloud
-    environment:s
+    environment:
       - CLOUDSDK_CONFIG=/tmp/.config/gcloud
       - CLOUDSDK_CORE_PROJECT


jobs/kpi-forecasting/kpi-forecasting/Utils/FitForecast.py
```diff
@@ -1,16 +1,18 @@
-from datetime import datetime
-import typing
```
perrymcmanis144, line 2: you should not be removing this import, see line 13

perrymcmanis144 (changes requested):
One nit.

I will say, as I am no longer the owner if you want to make this change then I will approve, but I would ask why this is actually desired behavior. As a probabilistic model, I would argue that adherence to correctness, even if conceptual, is more valuable than this level of repeatability.

But, if this makes your life easier, I will approve after resolving my one in-line

perrymcmanis144:
also @bochocki while you're making changes could you check this? https://github.com/mozilla/docker-etl/blob/main/jobs/kpi-forecasting/docker-compose.yml#L12

bochocki:
> also @bochocki while you're making changes could you check this? https://github.com/mozilla/docker-etl/blob/main/jobs/kpi-forecasting/docker-compose.yml#L12

Good catch! I thought I fixed that in my last PR (I had changed it locally), but I guess I didn't push the changes for that file.

jobs/kpi-forecasting/kpi-forecasting/Utils/FitForecast.py
bochocki: I didn't remove it :) This PR only alphabetized the imports and removed some extra line breaks; `import typing` is now on line 5 of the changed file.

bochocki:
> One nit.
> 
> I will say, as I am no longer the owner if you want to make this change then I will approve, but I would ask why this is actually desired behavior. As a probabilistic model, I would argue that adherence to correctness, even if conceptual, is more valuable than this level of repeatability.
> 
> But, if this makes your life easier, I will approve after resolving my one in-line

Good question! We sometimes need to re-run a forecast. For example, in KPI goal setting, we choose a single target value for DAU; it's important that this number is reproducible so that we can recapture it if we re-run the analysis at the end of the year. When the point estimates change between runs on the same underlying data, it makes life more difficult for us and for stakeholders.

I would argue that setting a random seed is not incorrect, even conceptually. There is _some_ underlying seed set right now, it's just not surfaced or recorded. As proposed, each run is still probabilistic; we're not defining what the outcome will be _a priori_, or limiting the space that the model has to explore its optimization. The uncertainty in the estimate is still captured in the confidence intervals. We just know that if we run the model with the same underlying data, we'll always get the same answer.

perrymcmanis144 (approved):

PR #121, opened 2023-06-02
Refactor KPI Forecasting

bochocki:
This is a major refactor that aims to make KPI forecasting easier to modify in the future. This refactor maintains the central design pattern -- forecasts are made from a metric based on configs set in a YAML file -- but most of the implementation has been refactored.

There is a related, ongoing project to [re-model the way we store forecast data in Big Query](https://docs.google.com/document/d/18esfJraogzUf1gbZv25vgXkHigCLefazyvzly9s-1k0/edit#heading=h.3vp2yzc6seen). One key change in this refactor is to format output data to be as close as possible to what we expect the new data model to look like, while also preserving the legacy data model for the time being. Methods related to legacy outputs are clearly marked in the code (usually including `legacy` in the method name) and are commented with `TODO` items that indicate which code paths can be removed once the transition has been made. I tested that the legacy data format was correct by copying legacy tables to `moz-fx-data-bq-data-science.bochocki` and appending output to those tables.

Key changes are:
  - Convert from a functional framework to an object-oriented framework. This will enable us to add new forecasting models more easily in the future by simply inheriting a `BaseForecast` class instead of writing bespoke logic for each new forecasting method.
  - Update the output to produce a single table that can accommodate different numbers of measurements via the `measures` column.
  - Use Metric Hub definitions instead of bespoke metric queries. By using versioned Metric Hub definitions, we can more easily ensure that the forecast is using up-to-date metric values.
 

Checklist for reviewer:

  responsibly.


jobs/kpi-forecasting/.gitignore
```diff
@@ -1,6 +1,12 @@
...
 .idea
+.ipython
+.ipynb_checkpoints
 .local
 .python-version
 .python_history
 .vscode
+
+# filetypes
+*.ipynb
```
kik-kik, line 12: Could you also add `*.pyc` to the ignore list?

jobs/kpi-forecasting/Dockerfile
```diff
@@ -12,6 +12,8 @@ RUN groupadd --gid ${USER_ID} ${GROUP_ID} && \
 
 WORKDIR ${HOME}
 
+RUN apt-get -y update
```
kik-kik, line 15: So the best practice would to combine commands into a single layer if possible.

In this case, we could do the following:
```
RUN apt-get -y update && apt-get -y install \
    git \
    && rm -rf /var/lib/apt/lists/*
```

Here's a Stackoverflow comment about this:
https://stackoverflow.com/a/39330458

jobs/kpi-forecasting/Dockerfile
```diff
@@ -12,6 +12,8 @@ RUN groupadd --gid ${USER_ID} ${GROUP_ID} && \
...
+RUN apt-get -y install git
 RUN pip install --upgrade pip
```
kik-kik, line 17: For the pip install commands you should also include the `--no-cache-dir` flag. In general in the Dockerfile we try to keep each layer as small as possible.

Can read more about Dockerfile best practices here: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/

jobs/kpi-forecasting/README.md
```diff
@@ -1,66 +1,85 @@
...
+### Local Python
 
-One of the challenges in automating multiple forecasts is that data formats may not always be consistent. This is a challenge precisely because Prophet demands that data conform to a very strict standard of two columns; 'ds' and 'y' as its primary inputs. Other columns can be added as regressors, but often these are trivial to implement in Python but challenging to implement in SQL, or vice versa, depending on their specifics.
+You can also run the code outside of a Docker container. The code below creates a new Conda environment called `kpi-forecasting-dev`.
+It assumes you have Conda installed. If you'd like to run the code in a Jupyter notebook, it is handy to install Jupyter in your `base` environment.
+The `ipykernel` commands below will ensure that the `kpi-forecasting-dev` environment is made available to Jupyter.
 
-Rather than try and force things to conform to an ultra-strict standard to solve a wide variety of forecasts, this repo focuses ONLY on KPIs, and is organized in such a way to make that easier to maintain.
+```sh
+conda create --name kpi-forecasting-dev python=3.10 pip ipykernel
```
kik-kik, line 47: Is the virtual environment created in the project directory by conda? If yes, perhaps we should include it in the `.gitignore`

jobs/kpi-forecasting/ci_job.yaml
```diff
@@ -11,5 +11,5 @@ build-job-kpi-forecasting:
         name: Build Docker image
         command: docker build -t app:build jobs/kpi-forecasting/
     - run:
-          name: Test Code
-          command: docker run app:build pytest --black
+        name: Test Code
+        command: docker run app:build pytest --black
```
kik-kik, line 15:
```suggestion
        command: docker run app:build black .
```

jobs/kpi-forecasting/kpi-forecasting/kpi_forecasting.py
```diff
@@ -1,66 +1,30 @@
...
+    "prophet": ProphetForecast,
+}
 
 
 def main() -> None:
-    args = get_args()
-    with open(args.config, "r") as config_stream:
-        config = yaml.safe_load(config_stream)
+    # Load the config
+    config = YAML(filepath=CLI().args.config).data
```
kik-kik, line 14: Is there a need to put YAML and CLI into separate modules? Both appear to be single functions and this appears like a quite a short module.

jobs/kpi-forecasting/kpi-forecasting/metric_hub.py
```diff
@@ -0,0 +1,101 @@
...
+            date the metric should be queried.
+        alias (str): An alias for the metric. For example, 'DAU' instead of
+            'daily_active_users'.
+        project (str): The Big Query project to use when establishing a connection
+            to the Big Query client.
+    """
+
+    app_name: str
+    slug: str
+    start_date: str
```
kik-kik, line 31: some validation on the date fields would be nice since this is user input and we expect a specific format.

jobs/kpi-forecasting/kpi-forecasting/metric_hub.py
```diff
@@ -0,0 +1,101 @@
...
+        if x is not None:
+            x = f"'{x}'"
+        return x
+
+    @property
+    def query(self) -> str:
+        """Build a string to query the relevant metric values from Big Query."""
+
+        # Make sure start_date and end_date strings are formatted correctly
+        start_date = self._enquote(self.start_date)
```
kik-kik, line 70: If the enquoting is only for the query then you could just include the `''` inside the query template instead and avoid having to do this additional operation.

jobs/kpi-forecasting/kpi-forecasting/models/prophet_forecast.py
```diff
@@ -0,0 +1,119 @@
...
+        if self.use_holidays:
+            self.model.add_country_holidays(country_name="US")
+
+        # Modify observed data to have column names that Prophet expects, and fit
+        # the model
+        self.model.fit(self.observed_df.rename(columns=self.column_names_map))
+
+    def _predict(self) -> pd.DataFrame:
+        """
+        Forecast using `self.model`.
```
kik-kik, line 37: Is this comment providing much value to the readers understanding?

jobs/kpi-forecasting/kpi-forecasting/models/prophet_forecast.py
```diff
@@ -0,0 +1,119 @@
...
+        `moz-fx-data-shared-prod.telemetry_derived.kpi_automated_forecast_v1`.
+        """
+        # TODO: This method should be removed once the forecasting data model is updated:
+        # https://docs.google.com/document/d/18esfJraogzUf1gbZv25vgXkHigCLefazyvzly9s-1k0.
+
+        df = self.model.predict(
+            self.dates_to_predict.rename(columns=self.column_names_map)
+        )
+
+        # set legacy column values
```
kik-kik, line 62: Comments should provide understanding not just state what the code does. In this case, it would be more valuable to understand why this is needed. What do we mean by legacy?

jobs/kpi-forecasting/kpi-forecasting/models/prophet_forecast.py
```diff
@@ -0,0 +1,119 @@
...
+            df["metric"] = "DAU"
+        else:
+            df["metric"] = self.metric_hub.alias
+
+        df["forecast_date"] = str(datetime.utcnow().date())
+        df["forecast_parameters"] = str(
+            json.dumps({**self.parameters, "holidays": self.use_holidays})
+        )
+
+        if "desktop" in self.metric_hub.alias.lower():
```
kik-kik, line 73: Is it possible for both `desktop` and `mobile` to be in the set?

jobs/kpi-forecasting/kpi-forecasting/models/prophet_forecast.py
```diff
@@ -0,0 +1,119 @@
...
+            "multiplicative_terms_lower",
+            "multiplicative_terms_upper",
+            "yhat",
+            "target",
+            "forecast_date",
+            "forecast_parameters",
+            "metric",
+        ]
+
+        for column in columns:
```
kik-kik, line 115: is this expected and intended behaviour? Or does one of these columns missing a sign of something bad?

jobs/kpi-forecasting/kpi-forecasting/pandas_extras.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def aggregate_to_period(
+    df: pd.DataFrame,
+    period: str,
+    aggregation: callable = np.sum,
+    date_col: str = "submission_date",
+) -> pd.DataFrame:
+    """Floor dates to the correct period and aggregate."""
+    if period.lower() in ["day", "month", "year"]:
```
kik-kik, line 22: Could avoid the if, else if you flipped around the condition `not in` and if True raise the exception. and if we make it through the block do your assignment. Looks like it should always happen.

jobs/kpi-forecasting/kpi-forecasting/pandas_extras.py
```diff
@@ -0,0 +1,29 @@
...
+        period = period[0]
+    else:
+        raise ValueError(f"Don't know how to floor dates by {period}.")
```
kik-kik, line 25: Is ValueError the appropriate Exception type for this error?

jobs/kpi-forecasting/requirements.txt
```diff
@@ -1,117 +1,75 @@
...
-executing==1.2.0
-Flask==2.3.2
-fonttools==4.39.3
-fs==2.4.16
-fugue==0.8.3
-fugue-sql-antlr==0.1.6
-google-api-core==2.11.0
-google-auth==2.17.3
-google-cloud-bigquery==3.10.0
+fonttools==4.40.0
```
kik-kik, line 16: Are all of these requirements necessary?

jobs/kpi-forecasting/kpi-forecasting/metric_hub.py
```diff
@@ -0,0 +1,101 @@
...
+        self.metric = metric
+        self.alias = self.alias or metric.name
+        self.submission_date_column = metric.data_source.submission_date_column
+
+        # Modify the metric source table string so that it formats nicely in the query.
+        self.from_expression = self.metric.data_source._from_expr.replace(
+            "\n", "\n" + " " * 19
+        )
+
+    def _enquote(self, x) -> Optional[str]:
```
kik-kik, line 51:
```suggestion
    def _enquote(self, x: str) -> Optional[str]:
```

jobs/kpi-forecasting/Dockerfile
bochocki: Thank you for the reference! I added `--no-cache-dir` to the pip commands.

jobs/kpi-forecasting/Dockerfile
bochocki: Thank you, it makes sense now that you point it out but I haven't had enough experience with building Docker containers to catch that. I've made this change.

jobs/kpi-forecasting/.gitignore
bochocki: done.

jobs/kpi-forecasting/README.md
bochocki: Done. By default conda adds environments to a common directory (`~/Users/username/conda/envs`) but it's easy enough to add `envs` to `.gitignore` in case we choose to store environments in the repo in the future.

jobs/kpi-forecasting/kpi-forecasting/kpi_forecasting.py
bochocki: That's a great question. I considered merging them, but I think that keeping them separate is helpful if the forecasting code is being run outside of a Docker context. For example, when developing locally in a Jupyter notebook, the code can instead look like:
```python
    config = YAML(filepath="local_config.yml").data
```

If it's okay, I'd prefer not to make this change for now.

jobs/kpi-forecasting/kpi-forecasting/metric_hub.py
bochocki: Good call. I'm now coercing these args to dates using `pd.to_datetime`. Pandas will undoubtedly catch more edge cases than I could write out.

jobs/kpi-forecasting/kpi-forecasting/metric_hub.py
bochocki: This is a great idea. I initially wrote this method because when `end_date is None` I was setting `end_string = "CURRENT_DATE()"`, and I didn't want that string to include quotes. But I'm now simply setting `end_date` to the current UTC date, so I can remove the `_enquote` method.

jobs/kpi-forecasting/kpi-forecasting/models/prophet_forecast.py
```diff
@@ -0,0 +1,119 @@
...
+
+        # Modify observed data to have column names that Prophet expects, and fit
+        # the model
+        self.model.fit(self.observed_df.rename(columns=self.column_names_map))
+
+    def _predict(self) -> pd.DataFrame:
+        """
+        Forecast using `self.model`.
+        """
+        # generate the forecast samples
```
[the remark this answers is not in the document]
bochocki: Nope! Removed.

jobs/kpi-forecasting/kpi-forecasting/models/prophet_forecast.py
bochocki: Removed.

jobs/kpi-forecasting/kpi-forecasting/models/prophet_forecast.py
```diff
@@ -0,0 +1,119 @@
...
+        samples = self.model.predictive_samples(
+            self.dates_to_predict.rename(columns=self.column_names_map)
+        )
+        df = pd.DataFrame(samples["yhat"])
+
+        # add dates_to_predict to the output
```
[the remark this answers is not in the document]
bochocki: Removed.

jobs/kpi-forecasting/kpi-forecasting/models/prophet_forecast.py
```diff
@@ -0,0 +1,119 @@
...
+        df["submission_date"] = self.dates_to_predict
+
+        return df
+
+    def _predict_legacy(self) -> pd.DataFrame:
+        """
+        Recreate the legacy format used in
+        `moz-fx-data-shared-prod.telemetry_derived.kpi_automated_forecast_v1`.
+        """
+        # TODO: This method should be removed once the forecasting data model is updated:
```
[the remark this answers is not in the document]
bochocki: There are several of these `TODO` throughout the code. They're still needed; there's a "legacy" BQ data model which will be deprecated once the new data model is in place. The code is mostly written to support what we expect the new data model to look like. But there are several methods in the code (usually of the format `_method_name_legacy`) that exist just to re-capture the legacy data format until the transition is complete.

jobs/kpi-forecasting/kpi-forecasting/models/prophet_forecast.py
bochocki: I made an effort to explain this in the PR description, but maybe that's not the best place for the explanation. I also added the following description to the readme:

> There is [an ongoing project](https://docs.google.com/document/d/18esfJraogzUf1gbZv25vgXkHigCLefazyvzly9s-1k0/edit) to re-model the way we store forecast data in Big Query. One key change in this refactor is to format output data to be as close as possible to what we expect the new data model to look like, while also preserving the legacy data model for the time being. Methods related to legacy outputs are clearly marked in the code (usually including legacy in the method name) and are commented with TODO items that indicate which code paths can be removed once the transition has been made.

jobs/kpi-forecasting/kpi-forecasting/models/prophet_forecast.py
bochocki: It is possible but unlikely for now. This code path will be deleted in the near-future when the new data model is complete.

I made the parsing more robust so that it handles the case that you mention.

jobs/kpi-forecasting/kpi-forecasting/models/prophet_forecast.py
bochocki: This is expected behavior in the "legacy" framework -- mobile and desktop forecasts returned different numbers of columns (example `"regressor_00"` [here](https://github.com/mozilla/docker-etl/blob/cdc52b62b4ef7150597733ba03829bf050b28e99/jobs/kpi-forecasting/kpi-forecasting/Utils/FitForecast.py#L33)). Most of these columns are not being used (hence the project to revamp the data model), but I left the logic in place for consistency during the transition.

jobs/kpi-forecasting/kpi-forecasting/pandas_extras.py
bochocki: I think it's okay. The [docs](https://docs.python.org/3/library/exceptions.html#ValueError) say:
> Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as [IndexError](https://docs.python.org/3/library/exceptions.html#IndexError).

jobs/kpi-forecasting/kpi-forecasting/pandas_extras.py
bochocki: 💯 clever! done.

jobs/kpi-forecasting/requirements.txt
bochocki: As far as I know they are... I started from a fresh environment and only installed the packages required to get the code running.

jobs/kpi-forecasting/kpi-forecasting/metric_hub.py
bochocki: I removed this method.

jobs/kpi-forecasting/ci_job.yaml
bochocki: I don't understand why, but the `build-docker-etl` check is still failing after updating from master and making this change.

jobs/kpi-forecasting/kpi-forecasting/models/prophet_forecast.py
kik-kik: If this is specific to the code, I'd include that information with the code. It's much easier to read and understand, I may look at the code without knowing that README has code specific comments.

jobs/kpi-forecasting/kpi-forecasting/pandas_extras.py
kik-kik: That's fair, I'd maybe make the message a little bit more user friendly and include in it what the expected values are.

jobs/kpi-forecasting/kpi-forecasting/models/prophet_forecast.py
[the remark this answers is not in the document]
kik-kik: That's okay, just wanted to make sure this is actually should be in the code. One suggestion I'd have, it sounds like there might be an issue related to this TODO, if that indeed is the case, could you include the ticket number with this comment to be able to find the status of this and additional context.

jobs/kpi-forecasting/kpi-forecasting/kpi_forecasting.py
kik-kik: of course this is fine. The modularity is good, was just unsure if really unnecessary with so little code.

kik-kik:
More general comment @bochocki, would it make sense to add unit tests for this code?

kik-kik:
@bochocki I made a commit to the branch to fix the `build-docker-etl` CI step. Basically the reason it was failing is that we needed to re-run `./script/update_ci_config` script to apply the changes to your `ci_job.yaml` definition to `.circleci/config.yml` file.

bochocki:
> More general comment @bochocki, would it make sense to add unit tests for this code?

Unit tests are definitely a good idea. I wonder if it'll be okay to add those in a follow-up PR? I'll need to mock data for some of the more useful tests and I'd like to think a bit more carefully about that before diving into an implementation.

kik-kik:
> > More general comment @bochocki, would it make sense to add unit tests for this code?
> 
> Unit tests are definitely a good idea. I wonder if it'll be okay to add those in a follow-up PR? I'll need to mock data for some of the more useful tests and I'd like to think a bit more carefully about that before diving into an implementation.

I'll leave that up to you to decide.

kik-kik (approved):
I assume you tested the code and it works as expected. I think it'd be great if we could add tests, but also won't block this change. Please resolve any open conversations prior to merging.

jobs/kpi-forecasting/kpi-forecasting/pandas_extras.py
bochocki: Done!

jobs/kpi-forecasting/kpi-forecasting/models/prophet_forecast.py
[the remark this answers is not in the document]
bochocki: I added a link to the Jira ticket instead of a link to the proposal doc.

jobs/kpi-forecasting/kpi-forecasting/models/prophet_forecast.py
bochocki: I removed this from the README in favor of including a link to the Jira ticket in the comments.

PR #168, opened 2024-01-25
Adjust MetricHub to take segments and WHERE clause

m-d-bowerman:
Adds functionality to `MetricHub` class to take

1. a SQL WHERE clause, which acts on the underlying data source
2. a dictionary of segments, to calculate the metric of interest GROUPED BY

Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/kpi-forecasting/kpi_forecasting/metric_hub.py
+++ b/jobs/kpi-forecasting/kpi_forecasting/metric_hub.py
@@ -5,7 +5,7 @@ from datetime import datetime
 from google.cloud import bigquery
 from mozanalysis.config import ConfigLoader
 from textwrap import dedent
-from typing import Optional
+from typing import Dict
 
 
 @dataclass
@@ -19,6 +19,11 @@ class MetricHub:
         slug (str): The Metric Hub slug for the metric.
         start_date (str): A 'YYYY-MM-DD' formatted-string that specifies the first
             date the metric should be queried.
+        segments (Dict): A dictionary of segments to use to group metric values.
+            The keys of the dictionary are aliases for the segment, and the
+            value is a SQL snippet that defines the segment.
+        where (str): A string specifying a condition to inject into a SQL WHERE clause,
+            to filter the data source.
         end_date (str): A 'YYYY-MM-DD' formatted-string that specifies the last
             date the metric should be queried.
         alias (str): An alias for the metric. For example, 'DAU' instead of
@@ -30,6 +35,8 @@ class MetricHub:
     app_name: str
     slug: str
     start_date: str
+    segments: Dict[str, str] = None
+    where: str = None
     end_date: str = None
     alias: str = None
     project: str = "mozdata"
@@ -52,8 +59,31 @@ class MetricHub:
             "\n", "\n" + " " * 19
         )
 
+        # Add query snippets for segments
+        if self.segments:
+            segment_select_query = []
+            for alias, sql in self.segments.items():
+                segment_select_query.append(f"  {sql} AS {alias},")
+            self.segment_select_query = "\n    ".join(segment_select_query)
+            self.segment_groupby_query = "\n ,".join(self.segments.keys())
+
+        self.where = f"AND {self.where}" if self.where else ""
+
     def query(self) -> str:
         """Build a string to query the relevant metric values from Big Query."""
+        if self.segments:
+            return dedent(
+                f"""
+                SELECT {self.submission_date_column} AS submission_date,
+                    {self.metric.select_expr} AS value,
+                    {self.segment_select_query}
+                FROM {self.from_expression}
+                WHERE {self.submission_date_column} BETWEEN '{self.start_date}' AND '{self.end_date}'
+                    {self.where}
+                GROUP BY {self.submission_date_column},
+                    {self.segment_groupby_query}
+            """
+            )
 
         return dedent(
             f"""
@@ -61,6 +91,7 @@ class MetricHub:
                    {self.metric.select_expr} AS value
               FROM {self.from_expression}
              WHERE {self.submission_date_column} BETWEEN '{self.start_date}' AND '{self.end_date}'
+                {self.where}
              GROUP BY {self.submission_date_column}
             """
         )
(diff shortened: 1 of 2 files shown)


jobs/kpi-forecasting/kpi_forecasting/metric_hub.py
```diff
@@ -52,15 +59,39 @@ def __post_init__(self) -> None:
...
+        self.where = f"AND {self.where}" if self.where else ""
+
     def query(self) -> str:
         """Build a string to query the relevant metric values from Big Query."""
+        if self.segments:
+            return dedent(
+                f"""
+                SELECT {self.submission_date_column} AS submission_date,
+                    {self.metric.select_expr} AS value,
+                    {self.segment_select_query}
```
jmsilverman, line 79: Not sure if this is 'more pythonic' or cleaner but with some edits above you can combine this query string into 1 version (i.e., not have the if statement on line 74.

To do that I think you'd need to 

- initialize `segment_select_query` as "" above line 63 and added a leading comma to it inside the line 63 if statement
- do the same thing for `segment_groupby_query`

jobs/kpi-forecasting/kpi_forecasting/tests/test_metric_hub.py
```diff
@@ -0,0 +1,43 @@
...
+        app_name="multi_product",
+        slug="mobile_daily_active_users_v1",
+        start_date="2024-01-01",
+    )
+    now = to_datetime(datetime.utcnow()).date()
+
+    query = test_metric_hub.query()
+    query_where = f"WHERE submission_date BETWEEN '2024-01-01' AND '{now}'\n GROUP BY"
+
+    assert query_where in query
```
jmsilverman, line 18: I love that you added tests!!
Would it be better to hardcode the entire expected query in order to make sure the full query is returned as expected?

jobs/kpi-forecasting/kpi_forecasting/tests/test_metric_hub.py
```diff
@@ -0,0 +1,43 @@
...
+        slug="mobile_daily_active_users_v1",
+        start_date="2024-01-01",
+        where="test_condition = condition",
+    )
+
+    query = test_metric_hub.query()
+    assert f"\n    AND {test_metric_hub.where}" in query
+
+
+def test_metrichub_with_segments():
```
jmsilverman, line 34: Is it worth adding a 4th test when using both 'where' *AND* 'segments'?

jmsilverman (changes requested):
Left a couple comments on the test and 1 comment in the main file about some possible refactoring.

jobs/kpi-forecasting/kpi_forecasting/tests/test_metric_hub.py
m-d-bowerman: I thought about that, too, and I don't think so, as this is the only piece of the query that would change if a WHERE clause is supplied. Also, the tests start looking really gross with all of those SQL strings hanging around.

jobs/kpi-forecasting/kpi_forecasting/tests/test_metric_hub.py
m-d-bowerman: Added

jobs/kpi-forecasting/kpi_forecasting/metric_hub.py
m-d-bowerman: Updated this, thanks!

jobs/kpi-forecasting/kpi_forecasting/metric_hub.py
```diff
@@ -52,17 +59,32 @@ def __post_init__(self) -> None:
...
+        # Add query snippets for segments
+        if self.segments:
+            segment_select_query = []
+            for alias, sql in self.segments.items():
+                segment_select_query.append(f"  {sql} AS {alias},")
+            self.segment_select_query = ",\n    ".join(segment_select_query)
+            self.segment_groupby_query = ",\n ,".join(self.segments.keys())
+        else:
+            self.segment_select_query = ""
+            self.segment_groupby_query = ""
```
jmsilverman, line 71: Another nit-picky pythonic question, should these 2 assignment go between lines 62 and 63 (i.e., outside the if/else) and then you can remove the 'else' altogether? I honestly don't know which is "more correct".

jmsilverman (approved):
Approving it! I left 1 last python style question, so take it or leave it!

jobs/kpi-forecasting/kpi_forecasting/metric_hub.py
```diff
@@ -52,17 +59,32 @@ def __post_init__(self) -> None:
             "\n", "\n" + " " * 19
         )
 
+        # Add query snippets for segments
+        if self.segments:
+            segment_select_query = []
+            for alias, sql in self.segments.items():
+                segment_select_query.append(f"  {sql} AS {alias},")
```
bochocki, line 66: Is the trailing comma here unnecessary? Maybe I'm not thinking through this correctly, but I imagine you're going to get a list like:
```python3
segment_select_query = ["A as a,", "B as b"]
```
and then when you use the `join` later the resulting string will be something like:

```SQL
A as a,,
B as b,
```

jobs/kpi-forecasting/kpi_forecasting/metric_hub.py
```diff
@@ -52,17 +59,32 @@ def __post_init__(self) -> None:
...
+            self.segment_select_query = ",\n    ".join(segment_select_query)
+            self.segment_groupby_query = ",\n ,".join(self.segments.keys())
```
bochocki, line 68: Similar question here. Won't `",\n ,"` give you two commas in a row for every segment? So the groupby query would look like:
```sql
,
, test_segment1,
, test_segment2
```

jobs/kpi-forecasting/kpi_forecasting/metric_hub.py
m-d-bowerman: You're right, thanks for catching that. I've changed it in the most recent commit, to remove the comma from the joined string.

jobs/kpi-forecasting/kpi_forecasting/metric_hub.py
m-d-bowerman: I used the `else` to avoid those variables being assigned twice for cases where the `if` condition is satisfied. I'm also not sure which is more correct, and I am curious if there is a settled preference for this (couldn't find anything in my brief googling about it).

PR #175, opened 2024-03-05
Add funnel forecasting class for search forecasting

m-d-bowerman:
Adds `funnel_forecast.py`, which contains the class to produce search funnel forecasts. Goal is to have this PR reviewed and merged by March 15th.

TODO:
  1. Add tests for new forecast class
  2. Check discrepancy between posterior median and components table `yhat`
  3. Set up example notebook with training, scoring, and validation

Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/kpi-forecasting/kpi_forecasting/configs/model_inputs/__init__.py
+++ b/jobs/kpi-forecasting/kpi_forecasting/configs/model_inputs/__init__.py
@@ -22,6 +22,7 @@ class ProphetRegressor:
     """
 
     name: str
+    description: str
     start_date: Optional[str] = None
     end_date: Optional[str] = None
     prior_scale: Union[int, float] = 1
--- a/jobs/kpi-forecasting/kpi_forecasting/configs/model_inputs/__init__.py
+++ b/jobs/kpi-forecasting/kpi_forecasting/configs/model_inputs/__init__.py
@@ -38,22 +38,3 @@ class ProphetHoliday:
     ds: List
     lower_window: int
     upper_window: int
-
-
-@attr.s(auto_attribs=True, frozen=False)
-class ModelConfig:
-
-    metric: str
-    slug: str
-    segment: Dict[str, str]
-    start_date: Optional[str] = None
-    holidays: Optional[pd.DataFrame] = None
-    regressors: Optional[List[ProphetRegressor]] = []
-    parameters: Optional[Dict[str, Union[List[float], float]]] = None
-    cv_settings: Optional[Dict[str, str]] = {
-        "initial": "365 days",
-        "period": "30 days",
-        "horizon": "30 days",
-        "parallel": "processes",
-    }
-    trend_change: Optional[str] = ""
--- a/jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
+++ b/jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
@@ -584,12 +584,12 @@ class FunnelForecast(BaseForecast):
                 self.components_df.dtypes == object
             ].index.tolist()
             self.components_df["metric_slug"] = self.metric_hub.slug
-            self.components_df["trained_at"] = self.trained_at
+            self.components_df["forecast_trained_at"] = self.trained_at
 
             schema = [
                 bigquery.SchemaField("submission_date", bq_types.DATE),
                 bigquery.SchemaField("metric_slug", bq_types.STRING),
-                bigquery.SchemaField("trained_at", bq_types.TIMESTAMP),
+                bigquery.SchemaField("forecast_trained_at", bq_types.TIMESTAMP),
             ]
             schema += [
                 bigquery.SchemaField(col, bq_types.STRING) for col in string_cols
--- a/jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
+++ b/jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
@@ -325,6 +325,9 @@ class FunnelForecast(BaseForecast):
 
             df_bias = df_cv.groupby("cutoff")[["yhat", "y"]].sum().reset_index()
             df_bias["pcnt_bias"] = df_bias["yhat"] / df_bias["y"] - 1
+            # Prophet splits the historical data when doing cross validation using
+            # cutoffs. The `.tail(3)` limits the periods we consider for the best
+            # parameters to the 3 most recent cutoff periods.
             bias.append(df_bias.tail(3)["pcnt_bias"].mean())
 
         min_abs_bias_index = np.argmin(np.abs(bias))
@@ -400,7 +403,9 @@ class FunnelForecast(BaseForecast):
         ]
 
         # join observed data to components df, which allows for calc of intra-sample
-        ## error rates and how components resulted in those predictions
+        # error rates and how components resulted in those predictions. The `fillna`
+        # call will fill the missing y values for forecasted dates, where only yhat
+        # is available.
         components_df = components_df.merge(
             segment_settings.segment_model.history[["ds", "y"]],
             on="ds",
--- a/jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
+++ b/jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
@@ -75,16 +75,12 @@ class FunnelForecast(BaseForecast):
 
         # Construct a DataFrame containing all combination of segment values
         ## in the observed_df
-        combination_df = (
-            self.observed_df[self.metric_hub.segments.keys()]
-            .drop_duplicates()
-            .reset_index()[self.metric_hub.segments.keys()]
-        )
+        combination_df = self.observed_df[
+            self.metric_hub.segments.keys()
+        ].drop_duplicates()
 
         # Construct dictionaries from those combinations
-        segment_combinations = []
-        for _, row in combination_df.iterrows():
-            segment_combinations.append(row.to_dict())
+        segment_combinations = combination_df.to_dict("records")
 
         # initialize a list to hold models for each segment
         ## populate the list with segments and parameters for the segment
@@ -97,6 +93,10 @@ class FunnelForecast(BaseForecast):
             model_params = getattr(
                 self.parameters["segment_settings"], segment[split_dim]
             )
+
+            holiday_list = []
+            regressor_list = []
+
             if model_params["holidays"]:
                 holiday_list = [
                     getattr(holiday_collection.data, h)
@@ -114,8 +114,8 @@ class FunnelForecast(BaseForecast):
                     segment=segment,
                     start_date=model_params["start_date"],
                     end_date=self.end_date,
-                    holidays=[] or [ProphetHoliday(**h) for h in holiday_list],
-                    regressors=[] or [ProphetRegressor(**r) for r in regressor_list],
+                    holidays=[ProphetHoliday(**h) for h in holiday_list],
+                    regressors=[ProphetRegressor(**r) for r in regressor_list],
                     grid_parameters=dict(model_params["grid_parameters"]),
                     cv_settings=dict(model_params["cv_settings"]),
                 )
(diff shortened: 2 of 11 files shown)


jobs/kpi-forecasting/kpi_forecasting/configs/model_inputs/__init__.py
```diff
@@ -38,22 +38,3 @@ class ProphetHoliday:
...
-    holidays: Optional[pd.DataFrame] = None
-    regressors: Optional[List[ProphetRegressor]] = []
-    parameters: Optional[Dict[str, Union[List[float], float]]] = None
-    cv_settings: Optional[Dict[str, str]] = {
-        "initial": "365 days",
-        "period": "30 days",
-        "horizon": "30 days",
-        "parallel": "processes",
-    }
-    trend_change: Optional[str] = ""
```
jmsilverman, line 59: I assume this is just unused code? We're sure removing it won't break any other models we have here (e.g., DAU Desktop and DAU Mobile)?

m-d-bowerman: Correct, this is a class that I added that's no longer being used, so there's no risk to removing it.

jaredsnyder:
General question:  Are all the modifications to the yaml files necessary for funnel_forecast or are they being made for another reason?  Could you provide some detail on why these changes are happening?

m-d-bowerman:
> General question: Are all the modifications to the yaml files necessary for funnel_forecast or are they being made for another reason? Could you provide some detail on why these changes are happening?

Yes, they're necessary to allow the class to still inherit from the `BaseForecast` class, which expects that `parameters` keyword. Because of that, the segment-specific model specifications had to be nested below that. I also changed the percentiles that it's using to summarize because I kept a wide format for the table, to simplify how it'll be used downstream.

jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
```diff
@@ -17,13 +32,590 @@ class SegmentModelHolder:
...
+            how="left",
+        ).fillna(0)
+
+        forecast_summarized = (
+            forecast_agg.set_index("submission_date")
+            # Add observed data samples to any overlapping forecasted period. This
+            # ensures that any forecast made partway through a period accounts for
+            # previously observed data within the period. For example, when a monthly
+            # forecast is generated in the middle of the month.
+            .add(overlap[["value"]].values)
```
jaredsnyder, line 426: Why is it being added here?  I don't follow the comment

jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
```diff
@@ -17,13 +32,590 @@ class SegmentModelHolder:
...
+        print(f"Forecasting from {self.start_date} to {self.end_date}.", flush=True)
+        self._set_seed()
+        self.predicted_at = datetime.utcnow()
+
+        for segment_settings in self.segment_models:
+            forecast_df = self._predict(segment_settings)
+            self._validate_forecast_df(forecast_df)
+
+            segment_settings.forecast_df = forecast_df.copy(deep=True)
+            del forecast_df
```
jaredsnyder, line 486: Why do you need to copy and manually delete here?

jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
```diff
@@ -17,13 +32,590 @@ class SegmentModelHolder:
...
+
+    def summarize(
+        self,
+        periods: List[str] = ["day", "month"],
+        numpy_aggregations: List[str] = ["mean"],
+        percentiles: List[int] = [10, 50, 90],
+    ) -> None:
```
jaredsnyder, line 493: docstring

jaredsnyder (changes requested):
Almost all the comments are documentation focused, I will do another pass after these are resolved and will focus more on functionality

jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
```diff
@@ -17,13 +32,590 @@ class SegmentModelHolder:
...
+        # build training dataframe
+        if task == "train":
+            df = (
+                self.observed_df.loc[
+                    (  # filter observed_df to rows that exactly match segment dict
+                        (
+                            self.observed_df[list(segment_settings.segment)]
+                            == pd.Series(segment_settings.segment)
+                        ).all(axis=1)
+                    )
```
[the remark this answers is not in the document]
m-d-bowerman: which piece of this are you suggesting should be its own variable? The index that matches from `.loc`?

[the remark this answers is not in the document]
jaredsnyder: Yeah, everything in lines 180-182 (though I guess leaving off the `all` might make sense if you have line length issues

jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
```diff
@@ -17,13 +32,590 @@ class SegmentModelHolder:
...
+        components_df = components_df.merge(
+            segment_settings.segment_model.history[["ds", "y"]],
+            on="ds",
+            how="left",
+        ).fillna(0)
+        components_df.rename(columns={"ds": "submission_date"}, inplace=True)
+
+        segment_settings.components_df = components_df.copy()
+
+        return df.loc[
```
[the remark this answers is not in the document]
jaredsnyder: Okay I was misunderstanding what line 337 does, think I'm good now

jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
```diff
@@ -17,13 +32,590 @@ class SegmentModelHolder:
...
+            print(
+                """
+                  Can only pass a list of length 3 as percentiles, for lower, mid, and upper values.
+                  Will default to using [10, 50, 90].
+                  """,
+                flush=True,
+            )
+            percentiles = [10, 50, 90]
+        # build a list of all functions that we'll summarize the data by
+        aggregations = [getattr(np, i) for i in numpy_aggregations]
```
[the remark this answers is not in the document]
m-d-bowerman: I don't see a way around doing this, unfortunately, as the aggregation functions are passed from the config files and loaded as strings, unless we add the `getattr` call to the YAML parser?

jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
m-d-bowerman: It's possible for the forecast and observed data to overlap when you aggregate, e.g. if you're forecasting for June and starting with today's date, your monthly aggregation would require those first few days of June from the historical data to be correct. This will take those days from the historical data and add them to the June forecast

jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
m-d-bowerman: Hmm, probably my paranoia. I'll remove it

jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
```diff
@@ -17,13 +32,590 @@ class SegmentModelHolder:
...
+        # aggregate metric to the correct date period (day, month, year)
+        observed_summarized = pdx.aggregate_to_period(
+            (
+                self.observed_df.loc[
+                    (
+                        (
+                            self.observed_df[list(segment_settings.segment)]
+                            == pd.Series(segment_settings.segment)
+                        ).all(axis=1)
+                    )
```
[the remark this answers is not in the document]
m-d-bowerman: Same comment as above

jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
jaredsnyder: Okay I think I get it.  Will the overlap be at most one period, or could you have a larger overlap and if so why does this still work?

jaredsnyder: And why do you need to additional aggregation step after this happens?

jaredsnyder (approved):
Code looks good, it is well documented and I can follow what is happening

jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
m-d-bowerman: I think the overlap could only be one period, yeah. I can't think of an instance where it wouldn't be. The aggregation after this will take those days that overlap include them when aggregating to the month.

jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
```diff
@@ -17,13 +32,704 @@ class SegmentModelHolder:
...
+
+        for params in param_grid:
+            m = self._build_model(segment_settings, params)
+            m.fit(test_dat)
+
+            df_cv = cross_validation(m, **segment_settings.cv_settings)
+
+            df_bias = df_cv.groupby("cutoff")[["yhat", "y"]].sum().reset_index()
+            df_bias["pcnt_bias"] = df_bias["yhat"] / df_bias["y"] - 1
+            bias.append(df_bias.tail(3)["pcnt_bias"].mean())
```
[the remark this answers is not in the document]
m-d-bowerman: Those cutoffs are the points in the historical data where forecasts are made when running cross validation. The `.tail(3)` limits the results to look at the best performance over the last 3 cutoffs, rather the whole time series. It looks like we could increase the `initial` parameter we pass to `cross_validation` to accomplish the same thing, but I'd have to dig a bit to be sure. Prophet documentation is garbage

jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
```diff
@@ -17,13 +32,704 @@ class SegmentModelHolder:
...
+            component_cols
+        ]
+
+        # join observed data to components df, which allows for calc of intra-sample
+        ## error rates and how components resulted in those predictions
+        components_df = components_df.merge(
+            segment_settings.segment_model.history[["ds", "y"]],
+            on="ds",
+            how="left",
+        ).fillna(0)
```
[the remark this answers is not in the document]
m-d-bowerman: Correct, I believe for the early parts of the series that are used to make initial estimates of trend and seasonality. I'll have to double check to be sure, though

jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
jaredsnyder: Is the second agg redundant?  Both `forecast_agg` and `observed_summarized` are aggregated to `submission_date`, right?

jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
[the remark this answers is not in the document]
jaredsnyder: Okay, maybe throw in a comment explaining the tail call if that doesn't work

jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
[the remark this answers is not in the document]
jaredsnyder: Leave a comment after you verify plz

jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
m-d-bowerman: They are, yeah, but this will re-perform the monthly aggregations. So when it adds in the overlap rows to the forecast set, from the historical set, that overlapping month will have 2 rows in the DataFrame. The second agg call aggregates that overlap.

jaredsnyder: You might be able to simplify it by concat-ing both unaggregated dataframes and then doing the agg?  But maybe that is harder to read?  idk if it's worth messing around with but this is one of the knottier bits of code in this module (which is indicative of how clearly the written the rest of it is!)

jaredsnyder: Sounds good to me, resolving

PR #183, opened 2024-04-04
Fixes #182 - Update webcompat-kb ETL job to fetch additional bugs and history

### LINKED ISSUE #182: Update webcompat-kb ETL job to fetch additional bugs and history
Overall import criteria - include all bugs that meet the following criteria:

- Product = ‘Web Compatibility’
- OR includes keyword(s) like ‘webcompat:’
- OR bugs which block a Web Compat knowledge base bug

New bug-level fields:

- Bug Creation Date
- Keywords
- URL
- User Story

Historical fields (non-bug level):

- Keyword addition and removal dates
- Status change dates

ksy36:
Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/webcompat-kb/webcompat_kb/main.py
+++ b/jobs/webcompat-kb/webcompat_kb/main.py
@@ -2,8 +2,10 @@ import click
 import logging
 import requests
 import re
+import time
 
 from google.cloud import bigquery
+from datetime import datetime, timezone
 
 BUGZILLA_API = "https://bugzilla.mozilla.org/rest"
 
@@ -23,11 +25,31 @@ FIELD_MAP = {
     "--": None,
 }
 
-
-def get_urls_from_story(user_story_str):
-    pattern = r"url:(\S*)"
-    return re.findall(pattern, user_story_str)
-
+FILTER_CONFIG = {
+    "wc": {
+        "product": "Web Compatibility",
+        "component": ["Knowledge Base", "Site Reports"],
+        "f1": "OP",
+        "f2": "bug_status",
+        "o2": "changedafter",
+        "v2": "2020-01-01",
+        "j1": "OR",
+        "f3": "resolution",
+        "o3": "isempty",
+        "f4": "CP",
+    },
+    "interventions": {
+        "product": "Web Compatibility",
+        "component": "Interventions",
+    },
+    "other": {
+        "v1": "Web Compatibility",
+        "f1": "product",
+        "o1": "notequals",
+        "keywords": "webcompat:",
+        "keywords_type": "regexp",
+    },
+}
 
 RELATION_CONFIG = {
     "core_bugs": {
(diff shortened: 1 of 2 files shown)


jobs/webcompat-kb/webcompat_kb/main.py
```diff
@@ -78,283 +100,543 @@ def get_urls_from_story(user_story_str):
...
-        if bug["id"] in bug_ids.get("core", []):
-            core_bugs.append(bug)
-        elif bug["id"] in bug_ids.get("breakage", []):
-            breakage_bugs.append(bug)
+        for bug_id in ids:
+            try:
+                logging.info(f"Fetching history from bugzilla for {bug_id}")
+                bug_history = self.fetch_history(bug_id, last_import_time)
+                bugs_history.append(bug_history)
+                time.sleep(2)
```
ksy36, line 449: This method would only fetch history for bugs changed since the last run to avoid calling the api unnecessarily.

I couldn't find any information about a rate limit for bugzilla api in the docs. Decided to add this wait in case there are a lot of bugs to fetch histories for at some point (for example, a mass update for all bugs in our component).

scholtzan (approved):

jobs/webcompat-kb/webcompat_kb/main.py
```diff
@@ -78,283 +100,543 @@ def get_urls_from_story(user_story_str):
...
+    return None
 
-    return result["bugs"]
 
+def parse_string_to_json(input_string):
+    if not input_string:
+        return ""
+
+    input_string = input_string.replace("\r\n", "\n")
+    lines = input_string.split("\n")
```
jgraham, line 125: I think `.splitlines()` would work here, but not a blocker.

jobs/webcompat-kb/webcompat_kb/main.py
```diff
@@ -78,283 +100,543 @@ def get_urls_from_story(user_story_str):
...
-        print(f"ERROR: {e}")
-        if job.errors:
-            for error in job.errors:
-                logging.error(error)
+        dep_ids = set(item for sublist in kb_dep_ids.values() for item in sublist)
 
-    table = client.get_table(bugs_table)
-    logging.info(f"Loaded {table.num_rows} rows into {table}")
+        # Check for missing bugs
+        all_ids = set([item["id"] for item in all_bugs])
```
jgraham, line 594: Not important, but you can write this as `{item["id"] for item in all_bugs}`

jgraham (approved):

jobs/webcompat-kb/webcompat_kb/main.py
ksy36: good point, thanks! I've changed it to use `.splitlines()`

jobs/webcompat-kb/webcompat_kb/main.py
```diff
@@ -78,283 +100,543 @@ def get_urls_from_story(user_story_str):
...
+        fetched_bugs = {}
+
+        for category, filter_config in FILTER_CONFIG.items():
+            logging.info(f"Fetching {category.replace('_', ' ').title()} bugs")
+            fetched_bugs[category] = self.fetch_bugs(filter_config)
+
+        kb_bugs = []
+        kb_depends_on_ids = set()
+
+        for bug in fetched_bugs["wc"]:
```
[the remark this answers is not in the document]
ksy36: The script only fetches blocking bugs of KB bugs (`depends_on` field) and doesn't fetch dependencies of Site reports, so this shouldn't be a problem.

```python
        for bug in fetched_bugs["wc"]:
            if bug["component"] == "Knowledge Base":
                kb_bugs.append(bug)
                kb_depends_on_ids.update(bug["depends_on"])
     
        core_bugs = self.fetch_bugs({"id": ",".join(map(str, kb_depends_on_ids))})

        all_bugs = (
            fetched_bugs["wc"]
            + fetched_bugs["interventions"]
            + fetched_bugs["other"]
            + core_bugs
        )
```
It made me think about a scenario where there could be duplicates - a core bug has `webcompat:` keyword and it's also a blocking bug for a KB entry (so it will be in both `fetched_bugs["other"]` and `core_bugs`). We don't have these kind of bugs right now, but it could be possible in the future.

For example, https://bugzilla.mozilla.org/show_bug.cgi?id=1796300 or https://bugzilla.mozilla.org/show_bug.cgi?id=1884638 both are core bugs with `webcompat:needs-diagnosis` with no kb entry (yet). Though in this case, we should probably remove the keyword once a bug is moved to core (i.e. diagnosed or can't be diagnosed further), or once kb entry is created.

Either way, I could deduplicate `all_bugs` by bug number before saving to avoid importing duplicates like that or similar :)

ksy36:
I think this is ready. We might want to add changes to history structure later on, but this should be good for now. 
Could you please merge it @scholtzan?

ksy36:
Thank you!

PR #189, opened 2024-04-29
Add new job for DAP Ads PPA Dev Collector

bbirdsong2:
This new job is to add a DAP collector for the ads eng team to test internally with the DAP infrastructure setup by the CTO office.

Checklist for reviewer:

  responsibly.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -123,6 +123,19 @@ jobs:
           name: Build Docker image
           command: docker build -t app:build jobs/dap-collector/
 
+  build-job-dap-collector-ppa-dev:
+    docker:
+      - image: << pipeline.parameters.git-image >>
+    steps:
+      - checkout
+      - compare-branch:
+          pattern: ^jobs/dap-collector-ppa-dev/
+      - setup_remote_docker:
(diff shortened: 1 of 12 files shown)


BenWu:
I triggered CI by pushing to the [add-dap-collector-ppa-dev-job](https://github.com/mozilla/docker-etl/tree/add-dap-collector-ppa-dev-job) branch but I'll let someone familiar with the business logic do a review

quiiver (commented):
I'm a little worried about the pattern we're introducing here. The implication is that every unique use case of DAP would require its own docker-etl container. I feel like there are a couple of patterns that could provide a more generic or reusable etl of DAP tasks. For example we could have a generic DAP collector task that takes some configuration for tasks, auth, etc. Then as part of the airflow DAG we transform the collected reports into whatever final output we care about. I'm happy to talk through the options we have at our disposal if you'd like.

That said I don't see any blockers to landing this as is.

quiiver (approved):

bbirdsong2:
@quiiver Yeah, I agree. DAP is still experimental so I am not currently bothered by the duplication, but once we decide to move forward with DAP I think we can better conslidate the infrastructure.

PR #198, opened 2024-05-22
Fixes #197 - Add translation step to the broken-site-report-ml job

### LINKED ISSUE #197: Add translation step to the broken-site-report-ml job
At this step the reports will be translated using ML.TRANSLATE and translated text will be used for classification. We also would like to store translation results in a separate table.

ksy36:
At this step the reports are translated using ML.TRANSLATE and translated text is used for classification and stored in a separate table.

Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/broken-site-report-ml/.flake8
+++ b/jobs/broken-site-report-ml/.flake8
@@ -1,2 +1,3 @@
 [flake8]
 max-line-length = 88
+ignore = E203
--- a/jobs/broken-site-report-ml/broken_site_report_ml/main.py
+++ b/jobs/broken-site-report-ml/broken_site_report_ml/main.py
@@ -111,6 +111,49 @@ def add_classification_results(client, bq_dataset_id, results):
     logging.info(f"Loaded {len(res)} rows into {table}")
 
 
+def save_translations(client, bq_dataset_id, results):
+    res = []
+    for uuid, result in results.items():
+        if not result["status"]:
+            bq_result = {
+                "report_uuid": uuid,
+                "translated_text": result["translated_text"],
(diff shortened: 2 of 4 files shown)


jobs/broken-site-report-ml/broken_site_report_ml/main.py
```diff
@@ -174,9 +225,64 @@ def get_missed_reports(client, last_run_time, bq_dataset_id):
...
+    """
+    query_job = client.query(query)
+    return list(query_job.result())
+
+
+def translate_reports(client, reports, bq_dataset_id):
+    result = {}
+    # Only translate reports that weren't translated
+    uuids_to_translate = [d["uuid"] for d in reports if not d["translated_text"]]
+
```
ksy36, line 265: If ML classification job fails due to bugbug not returning results in time, it will attempt to classify the same list of reports on the next run. In this case these reports will already be translated, so we don't need to translate them again and only query those without translation (`translated_text` is empty) by uuids.

jobs/broken-site-report-ml/.flake8
```diff
@@ -1,2 +1,3 @@
 [flake8]
 max-line-length = 88
+ignore = E203
```
ksy36, line 3: It looks like black and flake8 disagree on the whitespace in this function `yield data[i : i + size]`
I got `E203 whitespace before ':'` error when running flake8 after black autoformat, so added this ignore rule.

scholtzan (approved):

moztcampbell (approved):
Nice! This looks pretty clean for now.

ksy36:
Thanks!

PR #208, opened 2024-06-12
Fixes #207 - Add synthetic history records for keywords and fix a race condition

### LINKED ISSUE #207: Add synthetic history records for keywords and fix a race condition when some history records are not saved
We've noticed some entries are not recorded if a bug was edited during the time when the import is running.
In addition, bugs filed with keywords already set, don't have records about that in bugzilla, so we should create synthetic ones.

ksy36:
Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/webcompat-kb/webcompat_kb/main.py
+++ b/jobs/webcompat-kb/webcompat_kb/main.py
@@ -5,7 +5,7 @@ import re
 import time
 
 from google.cloud import bigquery
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 
 BUGZILLA_API = "https://bugzilla.mozilla.org/rest"
 
@@ -49,6 +49,19 @@ FILTER_CONFIG = {
         "o1": "notequals",
         "keywords": "webcompat:",
         "keywords_type": "regexp",
+    },
+    "parity": {
+        "f1": "OP",
+        "f2": "bug_status",
+        "o2": "changedafter",
+        "v2": "2020-01-01",
+        "j1": "OR",
+        "f3": "resolution",
+        "o3": "isempty",
+        "f4": "CP",
+        "f5": "keywords",
+        "o5": "regexp",
+        "v5": "parity-",
     },
 }
 
@@ -144,6 +157,10 @@ def parse_string_to_json(input_string):
     return result_dict
 
 
+def parse_datetime_str(s):
+    return datetime.strptime(s, "%Y-%m-%dT%H:%M:%SZ").replace(tzinfo=timezone.utc)
+
+
 class BugzillaToBigQuery:
     def __init__(self, bq_project_id, bq_dataset_id, bugzilla_api_key):
         self.client = bigquery.Client(project=bq_project_id)
@@ -151,6 +168,7 @@ class BugzillaToBigQuery:
         self.bugzilla_api_key = bugzilla_api_key
         self.bugs_fetch_completed = True
         self.history_fetch_completed = True
+        self.start_time = None
 
     def fetch_bugs(self, params=None):
         if params is None:
@@ -176,6 +194,7 @@ class BugzillaToBigQuery:
             "cf_last_resolved",
             "last_change_time",
             "whiteboard",
+            "creator",
         ]
 
         headers = {}
@@ -253,6 +272,7 @@ class BugzillaToBigQuery:
             fetched_bugs["wc"]
             + fetched_bugs["interventions"]
             + fetched_bugs["other"]
+            + fetched_bugs["parity"]
             + core_bugs
         )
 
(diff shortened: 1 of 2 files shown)


jobs/webcompat-kb/webcompat_kb/main.py
```diff
@@ -569,10 +591,70 @@ def update_history(self, records):
...
+
+        query = f"""
+                    SELECT *
+                    FROM `{self.bq_dataset_id}.bugs_history`
+                    WHERE number IN ({formatted_numbers})
+                """
+        result = self.client.query(query).result()
+        return result
+
+    def filter_only_unsaved_changes(self, history_updates, bug_ids):
```
ksy36, line 606: This is to avoid duplicating entries in case changes are picked up twice

jobs/webcompat-kb/webcompat_kb/main.py
```diff
@@ -590,30 +672,140 @@ def filter_bug_history_changes(self, history):
...
+        all_ids = {bug["id"] for bug in all_bugs}
+        new_ids = all_ids - existing_ids
+
+        logging.info(f"Fetching new bugs history: {list(new_ids)}")
+
+        new_bugs = [bug for bug in all_bugs if bug["id"] in new_ids]
+
+        history, _ = self.extract_relevant_changes(self.fetch_bugs_history(new_ids))
+
+        synthetic_history = self.create_synthetic_history(new_bugs, history)
```
ksy36, line 792: synthetic_history is only created for new bugs. (I did a re import from scratch and saved all history in `bugs_history_copy`, will rename it to `bugs_history` once this is merged)

scholtzan:
I granted you write permissions to the repo, so if you push changes to a branch to this repo instead of a fork it should kick of CI automatically

scholtzan (approved):

ksy36:
> I granted you write permissions to the repo, so if you push changes to a branch to this repo instead of a fork it should kick of CI automatically. And you should be able to merge things yourself too

Thank you!

jobs/webcompat-kb/webcompat_kb/main.py
```diff
@@ -590,30 +671,140 @@ def filter_bug_history_changes(self, history):
...
+                        )
+                    if change["removed"]:
+                        keyword_history[bug_id]["removed"].update(
+                            change["removed"].split(", ")
+                        )
+        return keyword_history
+
+    def create_synthetic_history(self, bugs, history):
+        bug_ids = set()
+        keyword_history = self.create_keyword_map(history)
```
jgraham, line 729: Unless we're using this somewhere else, it seems like what we want is a per-bug set of keywords for which there's either no added/removed entry or a removed entry before any added entry. In particular it seems like this approach is going to miss case where a keyword is added on creation, then removed, then readded.

jobs/webcompat-kb/webcompat_kb/main.py
```diff
@@ -569,10 +590,70 @@ def update_history(self, records):
...
+            return history_updates
 
-        for bug_history in history:
+        stored = defaultdict(list)
+
+        for row in existing_records:
+            row_dict = dict(row.items())
+            stored[row_dict["number"]].append(row_dict)
+
+        history_from_bq = dict(stored)
```
jgraham, line 618: Why do we copy this into a dict here?

jobs/webcompat-kb/webcompat_kb/main.py
```diff
@@ -569,10 +590,70 @@ def update_history(self, records):
...
+
+        history_bq_set = set()
+        for number, changes in history_from_bq.items():
```
jgraham, line 621: Seems like we don't ever actually use `number` inside the loop.

Can we in fact just construct a set of all updates in bigquery and a set of all updates in bugzilla and find the difference i.e. those changes not in bigquery but in bugzilla?

jgraham (changes requested):

ksy36:
I made the changes, could you please take another look, @jgraham?

jobs/webcompat-kb/webcompat_kb/main.py
```diff
@@ -590,30 +680,156 @@ def filter_bug_history_changes(self, history):
...
+        imported_ids = {bug["number"] for bug in rows}
+
+        return imported_ids
+
+    def update_keywords(self, keyword_records, change, timestamp):
+        for action in ["added", "removed"]:
+            keywords = change[action]
+            if keywords:
+                for keyword in keywords.split(", "):
+                    keyword_records[action][keyword] = timestamp
```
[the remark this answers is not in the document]
ksy36: I made a list of timestamps for each keyword and then changed the logic to sort the list by earliest timestamp. If the earliest in the list is "removed" for a given keyword, then we don't need to check whether "added" exists or when it happened, hopefully that simplifies the logic a little :)

jobs/webcompat-kb/webcompat_kb/main.py
```diff
@@ -857,7 +869,7 @@ def fetch_update_history(self, all_bugs):
         ) and self.history_fetch_completed:
             filtered_existing = self.filter_relevant_history(existing_bugs_history)
             filtered_records = filtered_existing + filtered_new_history
-            return self.update_history(filtered_records)
+            # return self.update_history(filtered_records)
```
jgraham, line 872: Need to re-enable all this commented out code :)

jgraham (commented):
Just need to revert the debugging changes and I think this will be good to go :)

jobs/webcompat-kb/webcompat_kb/main.py
ksy36: ah yes, thanks! 😄

jgraham (approved):

ksy36:
Moved this over to a branch that's not coming from a fork https://github.com/mozilla/docker-etl/pull/208

PR #213, opened 2024-06-21
[dap dev ppa collector] remove impression and click counts, adjust conversion counts offset

dmueller:
Checklist for reviewer:

[AE-364](https://mozilla-hub.atlassian.net/browse/AE-364)

This adjusts the ppa dev collector in preparation to creating the corresponding prod collector.

- clicks will not be reported via ppa
- impressions will not be reported via ppa
- adjust the offset of the vector given the other counts are not reported
- added a couple more data points to facilitate aggregations in the reporting
- adjusted the ads.json format to remove unused fields and organize fields that stayed

### testing
running locally while gloud authenticated to simulate the behavior when run by airflow. Currently it is running and emitting records into the `reports` table, but not to the `measurements` table.

After updating the version of janus collector being used here, then I was able to run this and see records get written to the dev tables.

- [X] Commits should reference a bug or github issue, if relevant (if a bug is referenced, the pull request should include the bug number in the title)
- [x] Scan the PR and verify that no changes (particularly to `.circleci/config.yml`) will cause environment variables (particularly credentials) to be exposed in test logs
- [x] Ensure the container image will be using permissions granted to [telemetry-airflow](https://github.com/mozilla/telemetry-airflow/) responsibly.

### DIFF
--- a/jobs/dap-collector-ppa-dev/dap_collector_ppa_dev/main.py
+++ b/jobs/dap-collector-ppa-dev/dap_collector_ppa_dev/main.py
@@ -20,8 +20,6 @@ ADS_SCHEMA = [
     bigquery.SchemaField("task_size", "INTEGER", mode="REQUIRED"),
     bigquery.SchemaField("task_id", "STRING", mode="REQUIRED"),
     bigquery.SchemaField("task_index", "INTEGER", mode="REQUIRED"),
-    bigquery.SchemaField("impression_count", "INTEGER", mode="REQUIRED"),
-    bigquery.SchemaField("click_count", "INTEGER", mode="REQUIRED"),
     bigquery.SchemaField("conversion_count", "INTEGER", mode="REQUIRED"),
 ]
 REPORT_SCHEMA = [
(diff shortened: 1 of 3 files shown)


jobs/dap-collector-ppa-dev/dap_collector_ppa_dev/main.py
```diff
@@ -7,7 +7,7 @@
 from google.cloud import bigquery
 import requests
 
-LEADER = "https://dap-07-1.api.divviup.org"
+LEADER = "https://dap-09-3.api.divviup.org"
```
dmueller, line 10: we've updated to 09-03, though when a task is created the leader & helpers it reports to are specified there. so if we have tasks created against multiple leaders then this will need to be adjusted

jobs/dap-collector-ppa-dev/dap_collector_ppa_dev/main.py
```diff
@@ -126,15 +127,16 @@ async def collect_once(task, timestamp, duration, hpke_private_key, auth_token):
                     if ad is not None:
                         cnt = {}
                         cnt["collection_time"] = collection_time
-                        cnt["placement_id"] = ad["debug"]["placementId"]
-                        cnt["ad_id"] = ad["debug"]["adId"]
-                        cnt["conversion_key"] = ad["debug"]["conversionKey"]
+                        cnt["placement_id"] = ad["advertiserInfo"]["placementId"]
```
dmueller, line 130: I adjusted the shape of the data in ads.json so that the job isn't reading a bunch of debug fields. The config is only used by this job and many of the fields exist to help enable similar aggregations as external reports.

gleonard-m (approved):
lgtm

cbguder (approved):

jobs/dap-collector-ppa-dev/dap_collector_ppa_dev/main.py
quiiver: Should this be configurable from outside the container?

quiiver (approved):

PR #218, opened 2024-06-26
Bug 1904928 - Add new image for exporting data for the Firefox-CI Tas…

ahal:
…kcluster instance

Checklist for reviewer:

  responsibly.


ahal:
I don't have permission to request review, but if you're looking for a good candidate to review the `fxci_etl` module, the `mozilla/releng` team is a good candidate.

I was debating putting the `fxci-etl` module into its own repo and then bootstrapping it here via pypi. I decided against it due to that requiring a multi-step deployment process.. But I could be convinced.

ahal:
telemetry-airflow PR that will use this image:
https://github.com/mozilla/telemetry-airflow/pull/2039

jobs/fxci-taskcluster-export/fxci_etl/loaders/bigquery.py
```diff
@@ -0,0 +1,67 @@
...
+        self.config = config
+        self._tables = {}
+
+        self.client = Client.from_service_account_info(
+            json.loads(
+                base64.b64decode(config.bigquery.credentials).decode(
+                    "utf8"
+                )
+            )
+        )
```
akkomar, line 41: I think we should rely on Airflow SA for write access to BQ, so this would become:
```suggestion
        self.client = Client()
```

akkomar (commented):
Generally this looks good, I'd only want to understand why we use custom credentials to write to BQ.

I added `releng` team to this repo so later on you can self-request reviews for this job.

jobs/fxci-taskcluster-export/fxci_etl/loaders/bigquery.py
ahal: Sounds good, I'll make the change. This is currently running in an EC2 instance in our `moz-fx-dev-releng` project, thus why I was using custom credentials.

jobs/fxci-taskcluster-export/fxci_etl/loaders/bigquery.py
```diff
@@ -0,0 +1,66 @@
...
+        tables = defaultdict(list)
+        for record in records:
+            tables[record.table_name()].append(record)
+
+        for name, rows in tables.items():
+            table = self.get_table(name)
+            errors = self.client.insert_rows(table, [asdict(row) for row in rows])
+
+            if errors:
+                pprint(errors, indent=2)
```
bhearsum, line 63: Should this be throwing an exception if there's errors? If not, do we just lose the rows? (The latter seems like it would be problematic if we have large swaths of failures.)

It appears that Pulse messages have already been ack'ed by the time this code runs.

bhearsum (commented):
This seems generally fine to me. I have one question about error handling I'd like answered before approving. (It's possible I'm just misunderstanding how things are run though.)

jobs/fxci-taskcluster-export/fxci_etl/loaders/bigquery.py
ahal: Oh yes.. thanks for the call out. I meant to fix that and forgot about it. I need to hook them up to the mechanism that saves the failed events in cloud storage.

ahal:
Hm, pushing to my fork doesn't seem to trigger the circleci builds.. I'm going to try re-opening this PR from the `ahal/fxci` branch on this repo.

PR #219, opened 2024-07-01
PPA Prod adjust how secrets are handled

dmueller:
https://mozilla-hub.atlassian.net/browse/AE-364

### problem
In the related airflow review, it was pointed out that we should handle the credentials differently there so that they aren't exposed in UIs. That switch shifts them from being ENV vars on the airflow job to be stored in google secrets manager.

From there, those secrets are passed to this job as environment variables rather than arguments.

### solution
compared to the xmatters setup to confirm my understanding
adjusted the hpke config and auth token to be read from os.environ instead. if these are missing then the requests to read from the collector will get error responses, which the job already captures
ran locally to confirm this adjustment
related airflow PR https://github.com/mozilla/telemetry-airflow/pull/2034

- handle secrets differently
- removed secret params from main method signature

Checklist for reviewer:

- [X] Commits should reference a bug or github issue, if relevant (if a bug is
- [X] Scan the PR and verify that no changes (particularly to
- [X] Ensure the container image will be using permissions granted to
  responsibly.

### DIFF
--- a/jobs/dap-collector-ppa-prod/dap_collector_ppa_prod/main.py
+++ b/jobs/dap-collector-ppa-prod/dap_collector_ppa_prod/main.py
@@ -2,6 +2,7 @@ import asyncio
 import click
 import datetime
 import math
+import os
 import time
 
 from google.cloud import bigquery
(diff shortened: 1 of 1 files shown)


quiiver (commented):
you can still use click for these https://click.palletsprojects.com/en/8.1.x/options/#values-from-environment-variables

quiiver (approved):

PR #223, opened 2024-07-01
Kpi forecasting: add validation

jaredsnyder:
Code to add validation table to the KPI Forecasting DAG

Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/kpi-forecasting/kpi_forecasting/validator.py
+++ b/jobs/kpi-forecasting/kpi_forecasting/validator.py
@@ -26,7 +26,12 @@ class Validator:
         self.output_table_id = (
             f"{self.output_project}.{self.output_dataset}.{self.output_table}"
         )
-        self.job_config = bigquery.QueryJobConfig(destination=self.output_table_id)
+
+        # table is so, so every time this runs it processes all the data
+        # and overwrites the old table
+        self.job_config = bigquery.QueryJobConfig(
+            destination=self.output_table_id, write_disposition="WRITE_TRUNCATE"
+        )
         self._load_config_data()
         self._extract_config_data()
 
--- a/jobs/kpi-forecasting/kpi_forecasting/validator.py
+++ b/jobs/kpi-forecasting/kpi_forecasting/validator.py
@@ -11,6 +11,7 @@ class Validator:
     output_project: str
     output_dataset: str
     output_table: str
+    input_config_path: str = "kpi_forecasting/configs"
 
     def __post_init__(self) -> None:
         """After initalization, set the following outputs after initialization
@@ -41,7 +42,7 @@ class Validator:
         are the values"""
         self.config_data = {}
         for config_file in self.input_config_list:
-            full_path = f"kpi_forecasting/configs/{config_file}"
+            full_path = f"{self.input_config_path}/{config_file}"
             config_data = YAML(full_path).data
             self.config_data[config_file] = config_data
 
--- a/jobs/kpi-forecasting/kpi_forecasting/validator.py
+++ b/jobs/kpi-forecasting/kpi_forecasting/validator.py
@@ -1,0 +1,147 @@
+from dataclasses import dataclass
+from google.cloud import bigquery
+
+from kpi_forecasting.inputs import YAML
+import pandas as pd
+
+
+@dataclass
+class Validator:
+    input_config_list: list[str]
+    output_project: str
+    output_dataset: str
+    output_table: str
+
+    def __post_init__(self) -> None:
+        """After initalization, set the following outputs after initialization
+        output_table_id: id for output table
+        job_config: used to set the output table when writing with bigquery
+        config_data: a dict of all the data from the list of configs provided
+            set inside of _load_config_data
+        input_table_full: id of the input data table extracted from the configs
+        dimension_list: indicates which columns of the input table represent
+            dimensions, where different combinations of values specify separate models
+            If a model has no such columns, set to an empty list
+        """
+        self.output_table_id = (
+            f"{self.output_project}.{self.output_dataset}.{self.output_table}"
+        )
+        self.job_config = bigquery.QueryJobConfig(destination=self.output_table_id)
+        self._load_config_data()
+        self._extract_config_data()
+
+    def _load_config_data(self):
+        """Extracts data from the list of config files passed to the class and stores it in the
+        config_data attribute. The filename is the key, and the contents (represnted as a DotMap)
+        are the values"""
+        self.config_data = {}
+        for config_file in self.input_config_list:
+            full_path = f"kpi_forecasting/configs/{config_file}"
+            config_data = YAML(full_path).data
+            self.config_data[config_file] = config_data
+
+    def _extract_config_data(self):
+        """Extracts data from the dictionary created by _load_config_data and uses it to set
+            the attributes below:
+                input_table_full: id of the input data table extracted from the configs
+                dimension_list: indicates which columns of the input table represent
+                    dimensions, where different combinations of values specify separate models
+                    If a model has no such columns, set to an empty list
+
+        Raises:
+            Exception: Raised if list of config files have different values for the dimension list
+            Exception: Raised if list of config files have different values for the input table
+        """
+        segment_data_list = []
+        input_table_list = []
+        config_file_list = list(self.config_data.keys())
+        for config_data in self.config_data.values():
+            # get segment data
+            metric_hub_data = config_data.metric_hub.toDict()
+            if "segments" in metric_hub_data:
+                segment_data = metric_hub_data["segments"]
+                segment_data_list.append(segment_data)
+            else:
+                segment_data_list.append(None)
+
+            # get input table info
+            input_table_list.append(config_data.write_results.toDict())
+
+        input_table_data = input_table_list.pop(0)
+        input_table_matches_first = [input_table_data == el for el in input_table_list]
+        if not all(input_table_matches_first):
+            config_file_list_string = " ".join(config_file_list)
+            raise Exception(
+                f"Input Table Data Does not all match for config list: {config_file_list_string}"
+            )
+
+        input_project = input_table_data["project"]
+        input_dataset = input_table_data["dataset"]
+        input_table = input_table_data["table"]
+
+        input_table_full = f"{input_project}.{input_dataset}.{input_table}"
+
+        segment_data = segment_data_list.pop(0)
+        segment_data_matches_first = [segment_data == el for el in segment_data_list]
+        if not all(segment_data_matches_first):
+            config_file_list_string = " ".join(config_file_list)
+            raise Exception(
+                f"Dimension Data Does not all match for config list: {config_file_list_string}"
+            )
+
+        if segment_data:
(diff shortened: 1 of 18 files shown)


jobs/kpi-forecasting/kpi_forecasting/validator.py
```diff
@@ -0,0 +1,153 @@
...
+        """Creates a query used to generate the validation table.
+
+        Returns:
+        (str): Query to generate validation table"""
+        if len(self.dimension_list) > 0:
+            segment_comma_separated = "," + ",".join(self.dimension_list)
+        else:
+            segment_comma_separated = ""
+        # in actual_deduped, the value for historical data won't change so we can use any_value without checking forecast_trained_at
+        compare_ctes = f"""WITH forecast_deduped as (SELECT submission_date, metric_alias {segment_comma_separated},
```
m-d-bowerman, line 118: I think we should pull this out into its own class. It'll be necessary to use a similar query for adjusting the forecasts, like for product change impacts, so it should be somewhere accessible outside of the `Validator` class.

jobs/kpi-forecasting/kpi_forecasting/validator.py
```diff
@@ -0,0 +1,153 @@
...
+                                                    MAX(forecast_trained_at) as forecast_trained_at,
+                                                    ANY_VALUE(value HAVING MAX forecast_trained_at) as forecast_value 
+                                                    FROM {self.input_table_full} 
+                                                    WHERE source='forecast' AND aggregation_period='day'
```
m-d-bowerman, line 122: Here, we might want to include validation for the `'month'` aggregation period, too. Would even be better to grab the `aggregation` value in the config `summarization` section to decide which aggregation periods should be validated over, then include `aggregation_period` as a dimension in the final validation table.

jobs/kpi-forecasting/kpi_forecasting/validator.py
m-d-bowerman: Although you are doing all of the metric calculation in the SQL. Are you opposed to making a general class to pull the data and using Python to calculate errors?

jobs/kpi-forecasting/kpi_forecasting/validator.py
```diff
@@ -0,0 +1,153 @@
...
+        return f"{compare_ctes} SELECT * FROM compare_data"
+
+    def fetch(self) -> pd.DataFrame:
+        """Uses the query produced by the query method to retrieve data
+            from bigquery and return as a pandas dataframe
+
+        Returns:
+            pd.DataFrame: Validation table as a pandas dataframe
+        """
+        query_job = bigquery.Client(project=self.output_project).query(self.query())
```
m-d-bowerman, line 144: I believe there's methods in the `QueryResult` (or whatever) that's returned by `Client.query` to create tables or write to tables, without needing to download the DataFrame, if there's no other data manipulation being done following the download.

m-d-bowerman (commented):
Left a few comments that we can discuss, especially around pulling the forecast data. I think we should have a more general class to do it, but I'm open to leaving it as-is, too, if you have strong opinions.

jobs/kpi-forecasting/kpi_forecasting/validator.py
jaredsnyder: Not opposed to doing the calculations in python.  Could you explain the other use case for the pull, so I can think through how to generalize it?

jobs/kpi-forecasting/kpi_forecasting/validator.py
jaredsnyder: I think that's what's currently happening in the `write` method.  I left a `fetch` method because I figured we'd want a way to get the dataframe

jobs/kpi-forecasting/kpi_forecasting/validator.py
jaredsnyder: I think it can be even simpler than that:  We know the aggregation period is written to the "aggergation_period" column, so we can just add that column to the groupby and join and we're done.  Pushed this change

jobs/kpi-forecasting/kpi_forecasting/validator.py
m-d-bowerman: The other case, currently, would be to pull forecast data to make post-hoc adjustments -- for instance, when product thinks a change will impact the forecasts by x amount, pull the forecasts and apply that x scalar.

jobs/kpi-forecasting/kpi_forecasting/validator.py
m-d-bowerman: Ahh, duh, I didn't even look closely at the write method after reading the fetch method. My b

jobs/kpi-forecasting/kpi_forecasting/configs/search_forecasting_model_performance.yaml
```diff
@@ -0,0 +1,6 @@
+input_config_list: ["search_forecasting_ad_clicks.yaml",  "search_forecasting_daily_active_users.yaml", "search_forecasting_search_count.yaml"]
```
m-d-bowerman, line 1: What are your thoughts on merging this with the current config files? I'd prefer having a `validation` section in each of those files

jaredsnyder: I think the strongest argument not to do so is that the validation/performance is at the table level, and many config files aggregate up to make a single table.  But really that might be an argument for agglomerating  the individual configs so we have table-level configs rather than against merging the model performance configs with the existing ones

jaredsnyder: If you like the idea of combining the existing config files so that each config file represents a single output table, maybe it makes sense to keep it as is for now and address the config files when we refactor?

m-d-bowerman: Yeah, I agree, the forecast config files should be combined into a single file for each data source. That makes sense, to keep it as-is until that config change is made.

m-d-bowerman (approved):

jobs/kpi-forecasting/kpi_forecasting/configs/search_forecasting_model_performance.yaml
```diff
@@ -0,0 +1,6 @@
...
+output_project: moz-fx-data-bq-data-science  # change before merging
+output_dataset: jsnyder # change before merging
+output_table: "search_validation" # change before merging
```
m-d-bowerman, line 4: Before merging, let's change the table name to include "forecast", e.g. "search_forecast_validation"

jobs/kpi-forecasting/kpi_forecasting/results_processing.py
```diff
@@ -0,0 +1,358 @@
...
+        job_config: used to set the output table when writing with bigquery
+        config_data: a dict of all the data from the list of configs provided
+            set inside of _load_config_data
+        input_table_full: id of the input data table extracted from the configs
+        dimension_list: indicates which columns of the input table represent
+            dimensions, where different combinations of values specify separate forecasts
+            If a forecast has no such columns, set to an empty list
+        """
+
+        # table is so, so every time this runs it processes all the data
```
jmsilverman, line 47: "table is so"? I don't get it.

jobs/kpi-forecasting/kpi_forecasting/results_processing.py
```diff
@@ -0,0 +1,358 @@
...
+            self.identifier_columns = (*self.identifier_columns, *self.dimension_list)
+
+        # need identifier columns to be a list to make it easy to do pandas operations later
+        self.identifier_columns = list(self.identifier_columns)
+
+    def _get_most_recent_forecasts(self, month_level_df: pd.DataFrame) -> pd.DataFrame:
+        """Adds the following columns to month_level_df:
+                - previous_forecast_month (timestamp):
+                    Timestamp of the first day of the month corresponding to the current forecast
+                - forecast_value_previous_month (float): forecast value for the previous montb
```
jmsilverman, line 154: montb -> month

jobs/kpi-forecasting/kpi_forecasting/tests/test_performance_analysis.py
```diff
@@ -0,0 +1,289 @@
+import pytest
```
jmsilverman, line 1: Should we have tests for `write` and/or `fetch`?

jmsilverman (commented):
Overall looks solid! The tests that exist are really good looking but I had a question about adding tests for the whole data table fetch&write process. I also think there might be some issues in the README file, but that's kinda low priority to fix here (as I mentioned in my comment in the code). And then there were some typos/confusion in comments. Otherwise, the meat of the code that's doing the comparisons and calculations looks pretty good to me!

jmsilverman (changes requested):
Officially requesting changes.

jobs/kpi-forecasting/README.md
```diff
@@ -65,14 +64,34 @@ cd ~/PATH/TO/CONDA/envs/kpi-forecasting-dev/lib/python3.10/site-packages/prophet
...
+
+For example, the following command forecasts Desktop DAU numbers:
+
+```sh
+python ~/kpi_forecasting.py -c ~/kpi_forecasting/configs/dau_desktop.yaml
+```
+
+Similarly, the model performance for the KPI forecast can be done by executing:
+```sh
+python ~/model_performance_analysis.py.py -c ~/kpi_forecasting/configs/kpi_model_validation.yaml
```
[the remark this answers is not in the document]
jaredsnyder: I left it as `performance_analysis` because it is also used for the search forecasting.

Yeah the README is kind of stale.  I did make sure the commands relevant to the PR worked but I'd say we check and update the rest of it in a separate PR.

jobs/kpi-forecasting/kpi_forecasting/results_processing.py
jaredsnyder: Should be "table is small" whoops

jobs/kpi-forecasting/kpi_forecasting/tests/test_performance_analysis.py
jaredsnyder: I'd argue that, in the absence of an easy way to mock querying BQ tables, it'd be more trouble than it's worth.  I did my best to put anything nontrivial that is happening outside of BQ in its own method so I could unit test that to minimize the amount in these functions.

jobs/kpi-forecasting/kpi_forecasting/results_processing.py
jaredsnyder: Fixed in latest push

jobs/kpi-forecasting/kpi_forecasting/results_processing.py
jaredsnyder: fixed in latest push

jmsilverman (approved):
LGTM!

PR #230, opened 2024-07-09
Miscellaneous fixes for the firefox-ci export job

ahal:
Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/fxci-taskcluster-export/fxci_etl/pulse/handler.py
+++ b/jobs/fxci-taskcluster-export/fxci_etl/pulse/handler.py
@@ -111,7 +111,9 @@ class BigQueryHandler(PulseHandler):
         if "workerId" in run:
             run_record["worker_id"] = run["workerId"]
 
-        self.records.append(Run.from_dict(run_record))
+        self.records.append(
+            Run.from_dict(self.config.bigquery.tables.runs, run_record)
+        )
 
         if data["runId"] == 0:
             # Only insert the task record for run 0 to avoid duplicate records.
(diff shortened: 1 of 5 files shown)


jobs/fxci-taskcluster-export/fxci_etl/pulse/handler.py
```diff
@@ -91,8 +91,14 @@ def __init__(self, config: Config, **kwargs: Any):
 
     def process_event(self, event):
         data = event.data
+
+        if data.get("runId") is None:
+            # This can happen if `deadline` was exceeded before a run could
```
bhearsum, line 96: To be clear: we can still have deadline exceeded with a `runId` (in which case we still process it, which is what we want)?

bhearsum (changes requested):
It looks like you need to drop the `table_name` methods in https://github.com/mozilla/docker-etl/blob/main/jobs/fxci-taskcluster-export/fxci_etl/pulse/records.py as well?

jobs/fxci-taskcluster-export/fxci_etl/pulse/handler.py
ahal: I'm not sure. I didn't come across this case in the wild, but I noticed that the TC docs mentioned that `runId` might not always be set and gave `deadline-exceeded` as an example of when that would happen.

For now, I think it's fine to ignore events where it's unset. Tasks that never ran don't have a cost.

ahal:
Thanks for catching those missed `table_name` methods!

jobs/fxci-taskcluster-export/fxci_etl/pulse/handler.py
bhearsum: I certainly agree that it's safe to ignore anything without a `runId`, for the record! This was mostly a note that `deadline-exceeded` may also happen when a `runId` is present, AIUI.

bhearsum (approved):

PR #232, opened 2024-07-09
refactored base_forecast and prophet_forecast to enable easier testing

jaredsnyder:
Checklist for reviewer:

  responsibly.


jobs/kpi-forecasting/kpi_forecasting/models/base_forecast.py
```diff
@@ -46,7 +45,7 @@ class BaseForecast:
 
     def __post_init__(self) -> None:
         # fetch observed observed data
-        self.collected_at = datetime.utcnow()
+        self.collected_at = datetime.now(timezone.utc)
```
bochocki, line 48: I think this changes `self.collected_at` to be timezone aware (e.g. `YYYY-MM-DD HH:MM:SS.000000+00:00`) instead of unaware (e.g. `YYYY-MM-DD HH:MM:SS.000000`). That seems reasonable, my only question is whether this impacts the write to BQ or the way the data will end up being stored.

jobs/kpi-forecasting/kpi_forecasting/models/base_forecast.py
```diff
@@ -70,26 +69,55 @@ def __post_init__(self) -> None:
             }
         )
 
-    def _fit(self) -> None:
-        """
-        Fit a forecasting model using `self.observed_df` that was generated using
-        Metric Hub data. This method should update `self.model`.
+    @abc.abstractmethod
+    def _fit(self, observed_df: pd.DataFrame) -> None:
```
bochocki, line 73: Not sure I follow why we're adding arguments to `_fit` and `_predict`. If the arguments are attributes of the class, why not just invoke the attributes in the method? This is what ends up happening in the `fit` and `predict` classes anyways.

I can see if we might want to generate the model and then fit/predict with different data or date ranges. In that case, I'd suggest modifying the arguments on the fit/predict methods so that they accept the same arguments, and default to the class's own attributes (see below).

jobs/kpi-forecasting/kpi_forecasting/models/base_forecast.py
```diff
@@ -133,205 +161,30 @@ def _validate_forecast_df(self) -> None:
...
-        return df
-
     def fit(self) -> None:
         """Fit a model using historic metric data provided by `metric_hub`."""
         print(f"Fitting {self.model_type} model.", flush=True)
         self._set_seed()
-        self.trained_at = datetime.utcnow()
-        self._fit()
+        self.trained_at = datetime.now(timezone.utc)
+        self._fit(self.observed_df)
```
bochocki, line 169:
```suggestion
        self._fit(observed_df)
```

jobs/kpi-forecasting/kpi_forecasting/models/base_forecast.py
```diff
@@ -133,205 +161,30 @@ def _validate_forecast_df(self) -> None:
...
-            "target",
-            "unit",
-            "forecast_parameters",
-            "forecast_date",
-        ]
-        df[string_cols] = df[string_cols].astype(str)
-
-        return df
-
     def fit(self) -> None:
```
bochocki, line 164:
```suggestion
    def fit(self, observed_df: pd.DataFrame = self.observed_df) -> None:
```

jobs/kpi-forecasting/kpi_forecasting/models/base_forecast.py
```diff
@@ -133,205 +161,30 @@ def _validate_forecast_df(self) -> None:
...
         """Fit a model using historic metric data provided by `metric_hub`."""
         print(f"Fitting {self.model_type} model.", flush=True)
         self._set_seed()
-        self.trained_at = datetime.utcnow()
-        self._fit()
+        self.trained_at = datetime.now(timezone.utc)
+        self._fit(self.observed_df)
 
     def predict(self) -> None:
```
bochocki, line 171:
```suggestion
    def predict(self, dates_to_predict: pd.DataFrame = self.dates_to_predict) -> None:
```

jobs/kpi-forecasting/kpi_forecasting/models/base_forecast.py
```diff
@@ -133,205 +161,30 @@ def _validate_forecast_df(self) -> None:
...
-        """Generate a forecast from `start_date` to `end_date`."""
+        """Generate a forecast from `start_date` to `end_date`.
+        Result is set to `self.forecast_df`"""
         print(f"Forecasting from {self.start_date} to {self.end_date}.", flush=True)
         self._set_seed()
-        self.predicted_at = datetime.utcnow()
-        self.forecast_df = self._predict()
+        self.predicted_at = datetime.now(timezone.utc)
+        self.forecast_df = self._predict(self.dates_to_predict)
```
bochocki, line 177:
```suggestion
        self.forecast_df = self._predict(dates_to_predict)
```

jobs/kpi-forecasting/kpi_forecasting/models/prophet_forecast.py
```diff
@@ -116,3 +120,267 @@ def _predict_legacy(self) -> pd.DataFrame:
                 df[column] = 0.0
 
         return df[columns]
+
+    def _summarize(
```
bochocki, line 124: This is a lot to look through and I don't want to miss something important... is this just copy-pasted from the old version of the base class, or are there changes as well?

jobs/kpi-forecasting/kpi_forecasting/models/base_forecast.py
jaredsnyder: Ah good catch, I changed it only because `utcnow()` is deprecated so let me see if there is a unaware replacement

jobs/kpi-forecasting/kpi_forecasting/models/prophet_forecast.py
jaredsnyder: Yeah everything is copied and pasted, the only changes are some of the stuff with the signatures we discussed above, and the `utcnow` thing

jobs/kpi-forecasting/kpi_forecasting/models/base_forecast.py
jaredsnyder: Yeah this is something I want to talk through because I keep going back and forth on it.  The idea is to "decouple" `BaseForecast` and it's children, so when you look at one of the child classes you minimize the stuff happening in the parent class you "need to know about".  I also thought it was a bit more intuitive having `_fit` be a function that you pass the training data to.  Finally, and this is more of a secondary consideration, it seemed a bit easier to test this way.

I like the idea of having the attributes be default args for `fit` and `predict`.  It does make the code more flexible!

jaredsnyder: I keep updating the second sentence of the comment because nothing is sounding quite right, but at the end of the day I guess it just felt cleaner and more readable to me?

jaredsnyder: Also whoops turns out you can't use self to set defaults like that

jobs/kpi-forecasting/kpi_forecasting/models/prophet_forecast.py
jaredsnyder: Followed the directions here and just removed the tzinfo: https://blog.miguelgrinberg.com/post/it-s-time-for-a-change-datetime-utcnow-is-now-deprecated

jobs/kpi-forecasting/kpi_forecasting/models/base_forecast.py
jaredsnyder: Followed the directions here and just removed the tzinfo: https://blog.miguelgrinberg.com/post/it-s-time-for-a-change-datetime-utcnow-is-now-deprecated

jobs/kpi-forecasting/kpi_forecasting/tests/test_prophet_forecast.py
```diff
@@ -0,0 +1,458 @@
...
+
+from kpi_forecasting.models.prophet_forecast import ProphetForecast
+
+
+def test_summarize_non_overlapping_day():
+    observed_start_date = "2124-01-01"
+    observed_end_date = "2124-02-01"
+
+    predict_start_date = "2124-02-02"
+    predict_end_date = "2124-03-01"
```
bochocki, line 14: nit -- This is simply picking dates far in the future, right? Sometimes I like to use the year 3000... "2124" is so specific that it could be confusing or misinterpreted as a typo. A comment could also help here.

bochocki (approved):
Thank you so much for creating the notebook to compare the results between this branch and main; that's a really helpful framework for ensuring that we're not doing anything that will break the forecasts. And these tests will make it easier to validate changes in the future.

Since the notebook doesn't actually run forecasts using each branch, it doesn't demonstrate whether this branch introduces breaking changes via the refactor the base class. But the changes are sensible and presumably both branches ran locally without errors.

After this is merged, let's trigger the Airflow task to make sure it runs in prod as well. In the off-chance that the task fails, we can revert the PR and debug.

Nice work!

jobs/kpi-forecasting/kpi_forecasting/tests/test_prophet_forecast.py
jaredsnyder: Good call, changed to 2100 (3000 raises an error, I think unix seconds get messed up sometime in the 2100s so maybe it's related to that?)

PR #237, opened 2024-07-12
feat(fxci): add logic to automatically create tables with correct sch…

ahal:
…emas if they don't exist

This ended up being a bit more complicated than I wanted, but since I had all the types of the records defined in the dataclass, I didn't want to hard code them all a second time just for a BigQuery schema. So instead this uses the `generate_schema` function to automatically derive the table schema from the `Record` classes.

Was it worth doing this way? Probably not. But it works well, so I think it's worth keeping. This code can be re-used for future tables as well.

Checklist for reviewer:

  responsibly.


jobs/fxci-taskcluster-export/fxci_etl/loaders/bigquery.py
```diff
@@ -71,12 +136,16 @@ def insert(self, records: list[Record] | Record):
...
+                table = obj.pop("table")
+                records.append(Record.from_dict(table, obj))
         except NotFound:
             pass
 
-        tables = defaultdict(list)
+        tables = {}
         for record in records:
+            if record.table not in tables:
+                self.ensure_table(record.table, record.__class__)
```
akkomar, line 147: How many rows are there per run usually?
Running this per row seems suboptimal. Could we do this once per each table that we have in the set of rows?

ahal: It's not super clear, but this is running one per table because on the next iteration `record.table` will be in `tables` and we won't enter this `if` block.

ahal: I agree it would be nice to just do this ensuring once at startup or something. But the `BigQueryLoader` class is designed to not know about any specific tables and I was trying to avoid hardcoding them.

jobs/fxci-taskcluster-export/fxci_etl/loaders/bigquery.py
```diff
@@ -1,23 +1,67 @@
...
+class BigQueryTypes(Enum, settings=NoAlias):  # type: ignore
+    DATE: TypeAlias = str
+    FLOAT: TypeAlias = float
+    INTEGER: TypeAlias = int
+    STRING: TypeAlias = str
+    TIMESTAMP: TypeAlias = int
+
+
+def generate_schema(cls):
+    assert is_dataclass(cls)
```
JohanLorenzo, line 34: I'm not familiar with the codebase so maybe it's common practice: I know Python assertions can easily be turned off by a flag. What's the reason for not raising an error here?

JohanLorenzo (approved):
I'm no expert in the way we create BigQuery tables so my review may not provide a lot of value. That being said, the patch makes sense to me 🙂

jobs/fxci-taskcluster-export/fxci_etl/loaders/bigquery.py
ahal: Yeah, we could.. Sometimes I use assertions almost like comments or type hints. In this case I didn't know how to do a type hint so figured an `assert` is good and clear enough. If assertions were disabled, this would raise an exception a little later on anyway (in the `fields(cls)` call).

PR #240, opened 2024-07-15
Webcompat metric history

jgraham:
Checklist for reviewer:

  responsibly.


jgraham:
This currently isn't working locally; I get a permissions error trying to query the current metric values. I think that's because we need to add some service account to the sheet underlying the `dim_bug_score` table, but I"m not sure which one.

jobs/webcompat-kb/webcompat_kb/main.py
```diff
@@ -1105,17 +1172,31 @@ def run(self) -> None:
...
-    bz_bq = BugzillaToBigQuery(bq_project_id, bq_dataset_id, bugzilla_api_key)
+def get_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--log-level",
+        choices=["debug", "info", "warn", "error"],
+        default="info",
+        help="Log level",
+    )
+    parser.add_argument("--bq-project-id", required=True, help="BigQuery project id")
```
ksy36, line 1186: should probably keep this with underscore as it's being passed in airflow https://github.com/mozilla/telemetry-airflow/blob/main/dags/webcompat_kb.py#L53

jobs/webcompat-kb/webcompat_kb/main.py
```diff
@@ -1024,6 +1024,73 @@ def fetch_update_history(self, all_bugs: BugsById) -> list[BugHistoryEntry]:
...
+
+        query = f"""
+                SELECT *
+                FROM `{metrics_table}`
+            """
+        rows = list(dict(row.items()) for row in self.client.query(query).result())
+        for row in rows:
+            row["recorded_date"] = today
+
+        job_config = bigquery.LoadJobConfig(
```
ksy36, line 1069: It fails with an error when serializing a date (and decimal) object at `load_table_from_json` with `TypeError: Object of type ... is not JSON serializable`. We can either serialize both, or just use `insert_rows` instead, as it converts objects to strings in the right format internally. 

I've tried this and it works:

```python
logging.info(f"Writing to {history_table} table")

table_ref = self.client.get_table(history_table)
errors = self.client.insert_rows(table_ref, rows)

if errors:
    logging.error(errors)
else:
    logging.info("Metrics history recorded")
    logging.info(f"Loaded {len(rows)} rows into {table_ref}")
```

ksy36 (approved):

ksy36:
@scholtzan wonder if you could you merge this, please? (so the image gets pushed to the registry)

PR #242, opened 2024-07-19
feat(kevel-metadata): adding new etl for extracting kevel metadata from kevel api into ads_derived table

braunreyes:
References https://mozilla-hub.atlassian.net/browse/AD-295

This is a new job that will extract ad data from the kevel API and load into a raw data table called `ads_derived.kevel_metadata_history_v1.

Checklist for reviewer:

- [x] Commits should reference a bug or github issue, if relevant (if a bug is
- [x] Scan the PR and verify that no changes (particularly to
- [x] Ensure the container image will be using permissions granted to
  responsibly.

TO DO:
Finalize the merge logic to comply with stakeholder expectations.  Right now the daily partition is overwritten on each run.  @sergiosonline @curtismorales you will want to discuss how you want to handle things like  'what if ad_id is in a different flight with different attributes in the same day?'.


curtismorales:
We took a different approach; this is no longer needed

PR #248, opened 2024-07-29
feat: kpi forecasting add funnel_forecast unit tests

jaredsnyder:
Adds unit tests for funnel_forecast and does some light refactoring

Checklist for reviewer:

  responsibly.


jaredsnyder:
Made a notebook to compare outputs.  For most of the rows (where the optimization lands on the same parameters), the difference is within 1% but there are some significant outliers.  Not sure if this is inevitable due to the stochastic nature of the prediction or not.  Notably we didn't have this issue with `prophet_forecast`

https://colab.research.google.com/drive/1NxD5eVwZ0Vw3UhZtZoIev5Q44qEEjSTJ#scrollTo=CjguJpSKsv7k

jaredsnyder:
The seed is set in both branches, so the results should be the same.  Investigating

jaredsnyder:
Validation notebook is updated and all tests pass.  Two changes were necessary:

- A bug we fixed was switching the values in `_add_regressor` so 1 was assigned when true and 0 when False, while in main it is the other way around.  This did change the values, usually by small amounts but occasionally by more.  It may be worth revisiting whether or not this change is worth it.  To check, I reverted this change to produce the `_branch_0s` tables
- there was another bug involving using the aggregate function on strings.  I updated the function and added some tests.

NB: [../eam-integrations/scripts/workday_everfi_integration.py](https://colab.research.google.com/drive/1NxD5eVwZ0Vw3UhZtZoIev5Q44qEEjSTJ#scrollTo=68HyVEYZ5C1Y)

jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
```diff
@@ -311,58 +389,53 @@ def _auto_tuning(self, segment_settings: SegmentModelSettings) -> Dict[str, floa
...
-            )
+            df[regressor.name] = (
+                (df["ds"] >= pd.to_datetime(regressor.start_date).date())
+                & (df["ds"] <= pd.to_datetime(regressor.end_date).date())
+            ).astype(int)
         return df
 
-    def _predict(self, segment_settings: SegmentModelSettings) -> pd.DataFrame:
+    def _predict(
+        self, dates_to_predict_raw: pd.DataFrame, segment_settings: SegmentModelSettings
```
m-d-bowerman, line 432: Was the purpose of passing the `dates_to_predict` df as an attribute to make testing simpler?

jobs/kpi-forecasting/kpi_forecasting/pandas_extras.py
```diff
@@ -26,4 +26,32 @@ def aggregate_to_period(
 
     x = df.copy(deep=True)
     x[date_col] = pd.to_datetime(x[date_col]).dt.to_period(period[0]).dt.to_timestamp()
-    return x.groupby(date_col).agg(aggregation).reset_index()
+
+    # treat numeric and string types separately
+    x_string = x.select_dtypes(include=["datetime64", object])
+    x_numeric = x.select_dtypes(include=["float", "int", "datetime64"])
+
+    if set(x_string.columns) | set(x_numeric.columns) != set(x.columns):
```
m-d-bowerman, line 34: Can you explain why this check is necessary? I can't think of an instance in either `ProphetForecast` or `FunnelForecast` that might cause this check to fail.

jobs/kpi-forecasting/kpi_forecasting/pandas_extras.py
```diff
@@ -26,4 +26,32 @@ def aggregate_to_period(
...
+        raise ValueError(
+            f"Columns do not have string or numeric type: {missing_columns_str}"
+        )
+
+    x_numeric_agg = x_numeric.groupby(date_col).agg(aggregation).reset_index()
+
+    # all values of x_string should be the same because it is just the dimensions
+    x_string_agg = x_string.drop_duplicates().reset_index(drop=True)
+
+    if len(x_string_agg) != len(x_numeric_agg):
```
m-d-bowerman, line 48: This checks that there is a single value on each aggregation date for each combination of segment dimensions, correct?

jobs/kpi-forecasting/kpi_forecasting/tests/test_funnel_forecast.py
```diff
@@ -0,0 +1,1603 @@
...
+    assert set(expected.columns) == set(output_df.columns)
+    pd.testing.assert_frame_equal(output_df, expected[output_df.columns])
+
+    # should not be any nulls outside the metric column
+    non_metric_columns = [
+        el
+        for el in output_df.columns
+        if el not in ["value", "value_low", "value_mid", "value_high"]
+    ]
+    assert not pd.isna(output_df[non_metric_columns]).any(axis=None)
```
m-d-bowerman, line 228: Why filter to `non_metric_columns` here? I don't think there should be null values in any of the columns.

m-d-bowerman (approved):
The changes to `funnel_forecast` look good. The test cases, too, look sufficient to catch errors in the methods that might arise from future changes to the classes. I left a couple of comments for my clarity, but let's go ahead with it rather than having this persist as a blocker.

jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
jaredsnyder: Yeah that's the main reason

jobs/kpi-forecasting/kpi_forecasting/pandas_extras.py
jaredsnyder: It's a fail-safe.  The idea is that this function doesn't cover all types so if an expected one appears it will throw an error instead of quietly doing something weird (There whole motivation for adding tests for this function was it essentially "quietly did something weird*" when I updated the code).  IE if a new column was boolean valued this would throw an error.

* The string-valued columns got aggregated like `+` so you'd see values like `desktopdesktopdesktopdesktopdesktop`, it was funny

jobs/kpi-forecasting/kpi_forecasting/pandas_extras.py
jaredsnyder: Right, I'm assuming that the string values are one-to-one with dimensions so they're unique and this essentially checks that and raises and error if that's not the case

jobs/kpi-forecasting/kpi_forecasting/tests/test_funnel_forecast.py
jaredsnyder: Currently for the observed data the percentile columns ("value_low", "value_mid", "value_high") are NaN.  Is that expected?

m-d-bowerman: Ahh, you right, whoops

m-d-bowerman:
> [Validation NB](https://colab.research.google.com/drive/1NxD5eVwZ0Vw3UhZtZoIev5Q44qEEjSTJ#scrollTo=68HyVEYZ5C1Y) is updated and all tests pass. Two changes were necessary:
> 
>     * A bug we fixed was switching the values in `_add_regressor` so 1 was assigned when true and 0 when False, while in main it is the other way around.  This did change the values, usually by small amounts but occasionally by more.  It may be worth revisiting whether or not this change is worth it.  To check, I reverted this change to produce the `_branch_0s` tables
> 
>     * there was another bug involving using the aggregate function on strings.  I updated the function and added some tests.

I added a bit to the notebook, checking for differences between values, and all differences were 0. The lil bit is right before the components section; have a look, make sure I didn't goof up the merge or sets I used?

jaredsnyder:
> > [Validation NB](https://colab.research.google.com/drive/1NxD5eVwZ0Vw3UhZtZoIev5Q44qEEjSTJ#scrollTo=68HyVEYZ5C1Y) is updated and all tests pass. Two changes were necessary:
> > ```
> > * A bug we fixed was switching the values in `_add_regressor` so 1 was assigned when true and 0 when False, while in main it is the other way around.  This did change the values, usually by small amounts but occasionally by more.  It may be worth revisiting whether or not this change is worth it.  To check, I reverted this change to produce the `_branch_0s` tables
> > 
> > * there was another bug involving using the aggregate function on strings.  I updated the function and added some tests.
> > ```
> 
> @jaredsnyder I added a bit to the notebook, checking for differences between values, and all differences were 0. The lil bit is right before the components section; have a look, make sure I didn't goof up the merge or sets I used?

@m-d-bowerman : that's because the tables you were looking at ran the data with that change reverted.  I loaded the tables with the change and made some plots to compare the differences here https://colab.research.google.com/drive/1NxD5eVwZ0Vw3UhZtZoIev5Q44qEEjSTJ#scrollTo=Look_at_distribution_of_differences_when_changing_1_to_0

m-d-bowerman:
Updates look good :shipit:  🦜

PR #253, opened 2024-08-02
Change fxci metric export to process all data from yesterday, rather than from 10 minutes ago

ahal:
Checklist for reviewer:

  responsibly.


bhearsum (approved):
This looks OK to me; I think it might be good to have a domain expert in our data/ETL review it as well though?

ahal:
It wouldn't hurt, but the impression I got is that they wanted us to self-serve this as much as possible.

ahal:
@akkomar hi, could you land this so it triggers the image push please?

Feel free to also review if you like, we're not entirely sure how hands on/off you and your team want to be for changes like this.

jobs/fxci-taskcluster-export/fxci_etl/metric/export.py
```diff
@@ -87,21 +88,32 @@ def get_timeseries(self, project: str, interval: TimeInterval):
...
+        """Return the time interval to query metrics over.
+
+        This will grab metrics all metrics from the last end time, up until
+        11:59:59 of yesterday. Ideally the metric export runs in a daily cron
+        task, such that it exports a days worth of data at a time.
+        """
+        utc = pytz.UTC
+        now = datetime.now(utc)
+        yesterday = now.date() - timedelta(days=1)
+        end_time = utc.localize(datetime.combine(yesterday, datetime.max.time()))
```
akkomar, line 100: The approach you're taking here is to handle execution date logic yourself. This will work, but we usually rely on a date parameter passed in from Airflow and make the job run idempotent (which IIUC is also the case here because you store the state in GCS).
To do this in a more standard way, I'd pass a `date` parameter from Airflow like in [this job](https://github.com/mozilla/telemetry-airflow/blob/main/dags/dap_collector.py#L64) and use it to process all data from that day. This way I'd be sure that `date` is a "complete" day and this would allow to simplify this function.

akkomar (commented):
> Feel free to also review if you like, we're not entirely sure how hands on/off you and your team want to be for changes like this.

We definitely can be hands on, review requests are welcome!

I added a comment about date handling logic inline. Relying on Airflow-provided execution date would be more idiomatic and probably easier to reason about during debugging. I am however fine with merging this as is if you prefer that.

akkomar:
We discussed this on Slack and decided to merge as-is for now, and @ahal will look into simplifying as described in https://github.com/mozilla/docker-etl/pull/253#discussion_r1705849802.

ahal:
Thanks for your help!

For posterity, here's the bug I filed to track the improvements:
https://bugzilla.mozilla.org/show_bug.cgi?id=1912067

Even more than the simplification, the ability to easily re-populate older days makes this worth implementing

PR #260, opened 2024-08-07
Add chunking to avoid long GET request to bugzilla

ksy36:
We pass bugzilla ids as a parameter to the GET request as a a string. There are more than 1k bugs and the length of this string started exceeding 8k characters last week. The server stopped returning anything for this one request and the import didn't proceed further (I think 8k is the limit). This PR chunks the request in 400 ids in each chunk.

Checklist for reviewer:

  responsibly.


jgraham (approved):
A good longer-term refactor might be to move the bugzilla logic into a helper class, but I think this is enough to fix the bug for now.

ksy36:
> A good longer-term refactor might be to move the bugzilla logic into a helper class, but I think this is enough to fix the bug for now.

Thanks! Yeah, that's a good idea.

ksy36:
@scholtzan could you merge this, please?

ksy36:
thank you :)

PR #263, opened 2024-08-09
Refactor Config files

jaredsnyder:
This PR modifies the config files, especially for the FunnelForecast jobs, to bring them more closely into alignment.  This is a first step toward adding the ability to fit different "sub-models" to different partitions of the data (`metric_hub.segments` set in the FunnelForecast config) to the base class.

DotMap was also removed because managing the conversion for dict to DotMap and back was beginning to be more trouble than it was worth as the config files got bigger and used in more complex ways.

Checklist for reviewer:

  responsibly.


jaredsnyder:
Question for @m-d-bowerman on this one: I'm trying to clean up/document all the start dates.  As part of that I added the `predict_historical_dates` parameter so that we wouldn't have to overwite `dates_to_predict` in the `__post_init__`.  I would like to use `predict_historical_dates` to modify the `start_date` attribute via `_default_start_date` rather than overwriting `dates_to_predict`.  The only place in FunnelForecast I can see this causing problems is here:

https://github.com/mozilla/docker-etl/blob/fd1435b74b2dee6edbdb55faafb5fc670b34c138/jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py#L494

What is the purpose of this line?  Also what do you think of the general approach?

jobs/kpi-forecasting/kpi_forecasting/models/base_forecast.py
```diff
@@ -40,7 +37,7 @@ class BaseForecast(abc.ABC):
     start_date: str
     end_date: str
     metric_hub: MetricHub
-    number_of_simulations: int = 1000
+    predict_historical_dates: bool = False
```
m-d-bowerman, line 40: Need to add a description of this arg to the docstring

m-d-bowerman:
> Question for @m-d-bowerman on this one: I'm trying to clean up/document all the start dates. As part of that I added the `predict_historical_dates` parameter so that we wouldn't have to overwite `dates_to_predict` in the `__post_init__`. I would like to use `predict_historical_dates` to modify the `start_date` attribute via `_default_start_date` rather than overwriting `dates_to_predict`. The only place in FunnelForecast I can see this causing problems is here:
> 
> https://github.com/mozilla/docker-etl/blob/fd1435b74b2dee6edbdb55faafb5fc670b34c138/jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py#L494
> 
> What is the purpose of this line? Also what do you think of the general approach?

In the `FunnelForecast` class, we also capture the intra-time series predictions, along with the Prophet components used to make that prediction, done here:
https://github.com/mozilla/docker-etl/blob/2f4b0231eda133133daaa5134d20de40152ca6de/jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py#L400

That's the motivation behind overwriting `dates_to_predict`, while retaining the `start_date` param to trim down the `forecast_df` to only dates in the future. I'll review the changes you made to see if we can remove the repeated application of that condition; I agree, it's incredibly clunky.

jaredsnyder:
> > Question for @m-d-bowerman on this one: I'm trying to clean up/document all the start dates. As part of that I added the `predict_historical_dates` parameter so that we wouldn't have to overwite `dates_to_predict` in the `__post_init__`. I would like to use `predict_historical_dates` to modify the `start_date` attribute via `_default_start_date` rather than overwriting `dates_to_predict`. The only place in FunnelForecast I can see this causing problems is here:
> > https://github.com/mozilla/docker-etl/blob/fd1435b74b2dee6edbdb55faafb5fc670b34c138/jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py#L494
> > 
> > What is the purpose of this line? Also what do you think of the general approach?
> 
> In the `FunnelForecast` class, we also capture the intra-time series predictions, along with the Prophet components used to make that prediction, done here:
> 
> https://github.com/mozilla/docker-etl/blob/2f4b0231eda133133daaa5134d20de40152ca6de/jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py#L400
> 
> That's the motivation behind overwriting `dates_to_predict`, while retaining the `start_date` param to trim down the `forecast_df` to only dates in the future. I'll review the changes you made to see if we can remove the repeated application of that condition; I agree, it's incredibly clunky.

Given that it gets filtered to dates in the future using `self.start_date` how is this different than just having `dates_to_predict` start at `self.start_date`?

m-d-bowerman:
> > > Question for @m-d-bowerman on this one: I'm trying to clean up/document all the start dates. As part of that I added the `predict_historical_dates` parameter so that we wouldn't have to overwite `dates_to_predict` in the `__post_init__`. I would like to use `predict_historical_dates` to modify the `start_date` attribute via `_default_start_date` rather than overwriting `dates_to_predict`. The only place in FunnelForecast I can see this causing problems is here:
> > > https://github.com/mozilla/docker-etl/blob/fd1435b74b2dee6edbdb55faafb5fc670b34c138/jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py#L494
> > > 
> > > What is the purpose of this line? Also what do you think of the general approach?
> > 
> > 
> > In the `FunnelForecast` class, we also capture the intra-time series predictions, along with the Prophet components used to make that prediction, done here:
> > https://github.com/mozilla/docker-etl/blob/2f4b0231eda133133daaa5134d20de40152ca6de/jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py#L400
> > 
> > That's the motivation behind overwriting `dates_to_predict`, while retaining the `start_date` param to trim down the `forecast_df` to only dates in the future. I'll review the changes you made to see if we can remove the repeated application of that condition; I agree, it's incredibly clunky.
> 
> Given that it gets filtered to dates in the future using `self.start_date` how is this different than just having `dates_to_predict` start at `self.start_date`?

The forecast dates get filtered that way, yeah, but the components table doesn't -- it'll have a row for every date in the observed data and the forecast dates.

jobs/kpi-forecasting/kpi_forecasting/models/base_forecast.py
jaredsnyder: Fixed

jaredsnyder:
I'm going to leave dealing with `start_date` for the next PR where I generalize the use of SegmentSettings objects.  So this one is really for a full review

jobs/kpi-forecasting/kpi_forecasting/results_processing.py
```diff
@@ -74,12 +74,13 @@ def _set_intra_forecast_agg_functions(self):
...
         """Extracts data from the list of config files passed to the class and stores it in the
-        config_data attribute. The filename is the key, and the contents (represnted as a DotMap)
+        config_data attribute. The filename is the key, and the contents
         are the values"""
         self.config_data = {}
         for config_file in self.input_config_list:
             full_path = f"{self.input_config_path}/{config_file}"
-            config_data = YAML(full_path).data
+            config_data = load_yaml(full_path)
+            print(config_data)
```
bochocki, line 83: Is this print statement leftover from debugging? Or do we want to print that to, e.g. the Airflow logs?

bochocki (commented):
This looks good to me, although I'll defer to @m-d-bowerman for the approval since there are some funnel forecast changes that I'm not familiar with.

jobs/kpi-forecasting/kpi_forecasting/results_processing.py
jaredsnyder: whoops that should be taken out, fixed

jaredsnyder:
Created a notebook to validate that the outputs are the same.  Ended up finding an error that was corrected in the latest commit, but now everything matches

https://colab.research.google.com/drive/1dLeLUz_99ln9PC1AG-izZILj9-zIJHmJ#scrollTo=oukMs3d3qiEJ

m-d-bowerman (approved):

PR #272, opened 2024-08-23
Implement SKLearn interface

jaredsnyder:
Changes:
- data pull via metric club removed from all classes
- start_date and end_date attributes removed from all classes, time filtering will now happen in `kpi_forecasting.py` before passing data to model classes
- New class, `BaseEnsembleForecast`, created to deal with segmented models like FunnelForecast used to implement
- New class, `ProphetAutotunerForecast`,  created to implement automated hyperparameter tuning
- `FunnelForecast` recreated as a `BaseEnsembleForecast` that uses a `ProphetAutotunerForecast` as the base model
- `summarize` and `write_functions`, along with all the functions called within them, moved outside of classes

Checklist for reviewer:

  responsibly.


jaredsnyder:
Here's a notebook to validate the PR.  We're not getting an exact match on the search forecasts but @m-d-bowerman and I have concluded the models match and the difference is due to how prophet sets the seed: https://colab.research.google.com/drive/1dLeLUz_99ln9PC1AG-izZILj9-zIJHmJ#scrollTo=70upJ3eUTvkh

jaredsnyder:
Note on the validation: https://docs.google.com/document/d/1kG75iCFHSxBYVz6EcaYhozOZ9KfK7ncKvB5YfOmaB6I/edit?usp=sharing

jobs/kpi-forecasting/kpi_forecasting/configs/dau_mobile.yaml
```diff
@@ -7,26 +7,26 @@ metric_hub:
...
-  dataset: "telemetry_derived"
-  dataset_legacy: "telemetry_derived"
-  table: "kpi_forecasts_v0"
+  project: "moz-fx-data-bq-data-science"
+  project_legacy: "moz-fx-data-bq-data-science"
+  dataset: "jsnyder"
+  dataset_legacy: "jsnyder"
+  table: "kpi_forecasts_v0_branch"
+  forecast_table_legacy: "kpi_automated_forecast_v1_branch"
+  confidences_table_legacy: "kpi_automated_forecast_confidences_v1_branch"
```
bochocki, line 32: Are these just set for testing purposes?

jobs/kpi-forecasting/kpi_forecasting.py
```diff
@@ -11,22 +25,157 @@
...
+    def __init__(self, config_path):
+        self.config_data = load_yaml(filepath=config_path)
+        if isinstance(self.config_data["forecast_model"]["parameters"], list):
+            self.model_type = "funnel"
+            self.model_class = FunnelForecast
+            self.segments = list(self.config_data["metric_hub"]["segments"].keys())
+        else:
+            self.model_type = "prophet"
+            self.model_class = ProphetForecast
+            self.segments = None
```
bochocki, line 38: This logic seems quite brittle; it doesn't really leave room for a forecast model type that has a list of parameters and is _not_ FunnelForecast. Why not leave this as a parameter in the config?

jobs/kpi-forecasting/kpi_forecasting/configs/dau_mobile.yaml
jaredsnyder: oh god don't want to go through this whole thing again, good catch, pushed a fix

jobs/kpi-forecasting/kpi_forecasting.py
jaredsnyder: Yeah that's fair, added model_type back to the configs and updated this logic

jobs/kpi-forecasting/kpi_forecasting.py
```diff
@@ -11,22 +25,160 @@
...
+
+    def fit(self, observed_df):
+        # the model parameters are mixed in a bit with paraters to configure
+        # the predict dates so we have to do this
+
+        if self.model_type == "funnel":
+            model_parameters = {
+                "parameters": self.config_data["forecast_model"]["parameters"]
+            }
+            model_parameters["segments"] = self.segments
```
bochocki, line 91:
```suggestion
            model_parameters = {
                "parameters": self.config_data["forecast_model"]["parameters"],
                "segments": self.segments,
            }
```

jobs/kpi-forecasting/kpi_forecasting.py
```diff
@@ -11,22 +25,160 @@
...
+        start_date = pd.to_datetime(
+            self.config_data["forecast_model"]["forecast_start"]
+            or self._default_start_date(observed_df)
+        )
+        self.start_date = start_date
+        end_date = pd.to_datetime(
+            self.config_data["forecast_model"]["forecast_end"]
+            or self._default_end_date()
+        )
+        self.end_date = end_date
```
bochocki, line 78: Maybe there's a good reason not to do this; it's a bit of "code golf", but imo it's also slightly easier to scan. Happy to go either way.
```suggestion
        self.start_date = pd.to_datetime(
            self.config_data["forecast_model"]["forecast_start"]
            or self._default_start_date(observed_df)
        )
        self.end_date = pd.to_datetime(
            self.config_data["forecast_model"]["forecast_end"]
            or self._default_end_date()
        )
```

jobs/kpi-forecasting/kpi_forecasting.py
```diff
@@ -11,22 +25,160 @@
...
+        summary_df["forecast_predicted_at"] = self.predicted_at
+
+        return summary_df
+
+    def get_raw_data(self):
+        metric_hub = MetricHub(**self.config_data["metric_hub"])
+        self.collected_at = datetime.now(timezone.utc).replace(tzinfo=None)
+        observed_df = metric_hub.fetch()
+        # set attribute to generate metadata later
+        self.metric_hub = metric_hub
```
bochocki, line 65:
```suggestion
        self.metric_hub = MetricHub(**self.config_data["metric_hub"])
        self.collected_at = datetime.now(timezone.utc).replace(tzinfo=None)
        observed_df = self.metric_hub.fetch()
```

jobs/kpi-forecasting/kpi_forecasting.py
```diff
@@ -11,22 +25,160 @@
...
+        if model_type == "funnel":
+            self.model_type = "funnel"
+            self.model_class = FunnelForecast
+            self.segments = list(self.config_data["metric_hub"]["segments"].keys())
+        elif model_type == "prophet":
+            self.model_type = "prophet"
+            self.model_class = ProphetForecast
+            self.segments = None
+        else:
+            raise ValueError(f"Model type {model_type} is not supported")
```
bochocki, line 41:
```suggestion
        self.model_type = self.config_data["model_type"]
        if self.model_type == "funnel":
            self.model_class = FunnelForecast
            self.segments = list(self.config_data["metric_hub"]["segments"].keys())
        elif self.model_type == "prophet":
            self.model_class = ProphetForecast
            self.segments = None
        else:
            raise ValueError(f"Model type {self.model_type} is not supported")
```

jobs/kpi-forecasting/kpi_forecasting.py
```diff
@@ -11,22 +25,160 @@
...
+            model_parameters["segments"] = self.segments
+        elif self.model_type == "prophet":
+            model_parameters = self.config_data["forecast_model"]["parameters"]
+        model = self.model_class(**model_parameters)
+        self.trained_at = datetime.now(timezone.utc).replace(tzinfo=None)
+        return model.fit(observed_df)
+
+    def predict_and_summarize(self, model, predict_dates, observed_df):
+        raw_predictions = model.predict(predict_dates)
+        self.predicted_at = datetime.now(timezone.utc).replace(tzinfo=None)
```
bochocki, line 100: Just to be consistent with the `fit` class, which sets the timestamp immediately before fitting.
```suggestion
        self.predicted_at = datetime.now(timezone.utc).replace(tzinfo=None)
        raw_predictions = model.predict(predict_dates)
```

jobs/kpi-forecasting/kpi_forecasting.py
```diff
@@ -11,22 +25,160 @@
...
+
+        # add forecast model metadata columns
+        summary_df["forecast_start_date"] = self.start_date
+        summary_df["forecast_end_date"] = self.end_date
+        summary_df["forecast_trained_at"] = self.trained_at
+        summary_df["forecast_predicted_at"] = self.predicted_at
+
+        return summary_df
+
+    def get_raw_data(self):
```
bochocki, line 60: I think "raw" may be too ambiguous here.
```suggestion
    def get_historical_data(self):
```

jobs/kpi-forecasting/kpi_forecasting.py
```diff
@@ -11,22 +25,160 @@
...
-    if model_type in MODELS:
-        metric_hub = MetricHub(**config["metric_hub"])
-        model = MODELS[model_type](metric_hub=metric_hub, **config["forecast_model"])
+    pipeline = KPIPipeline(config_path)
 
-        model.fit()
-        model.predict()
-        model.summarize(**config["summarize"])
-        model.write_results(**config["write_results"])
+    observed_df = pipeline.get_raw_data()
```
bochocki, line 172:
```suggestion
    observed_df = pipeline.get_historical_data()
```

jobs/kpi-forecasting/kpi_forecasting/models/base_forecast.py
```diff
@@ -115,116 +59,206 @@ def _validate_forecast_df(self, forecast_df: pd.DataFrame) -> None:
...
+            model = segment_model["model"]
+            model._set_seed()
+            predict_df = model.predict(dates_to_predict_segment)
+
+            # add segments on as columns
+            for column, value in segment_model["segment"].items():
+                predict_df[column] = value
+            predict_df["forecast_parameters"] = json.dumps(model._get_parameters())
+            segment_model["forecast"] = predict_df
+        self.forecast_list = [el["forecast"] for el in self.segment_models]
```
bochocki, line 263: Why not use `i` or `x` here? "el" seems nonstandard and confusing.
```suggestion
        self.forecast_list = [i["forecast"] for i in self.segment_models]
```

jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
```diff
@@ -1,370 +1,48 @@
...
         # Prophet splits the historical data when doing cross validation using
         # cutoffs. The `.tail(3)` limits the periods we consider for the best
         # parameters to the 3 most recent cutoff periods.
-        return df_bias.tail(3)["pcnt_bias"].mean()
+        return np.abs(df_bias.tail(3)["pcnt_bias"].mean())
 
-    def _auto_tuning(
-        self, observed_df, segment_settings: SegmentModelSettings
-    ) -> Dict[str, float]:
+    def _auto_tuning(self, observed_df) -> ProphetForecast:
```
bochocki, line 45: This is a useful idea. My favorite model tuning fact is that random search is usually more efficient than grid search, especially across higher-dimensional spaces. The KPI Prophet models have been tuned with random search. If this method is computationally intensive or long-running, switching to random search could be a quick improvement.

- [A concise explanation on Stack Exchange](https://stats.stackexchange.com/a/209409)
- [Figure 11.2 in _Deep Learning_](https://www.deeplearningbook.org/contents/guidelines.html) 
image

jobs/kpi-forecasting/kpi_forecasting/models/prophet_forecast.py
```diff
@@ -3,65 +3,339 @@
...
+
+    def __post_init__(self):
+        holiday_list = []
+        regressor_list = []
+
+        if self.holidays == []:
+            self.holidays = None
+            self.holidays_raw = None
+        elif not self.holidays:
+            self.holidays_raw = None
```
bochocki, line 132: What are these different conditionals checking for? The first one is clearly checking for an empty list. Is the second one checking if self.holidays is None?

jobs/kpi-forecasting/kpi_forecasting/models/prophet_forecast.py
```diff
@@ -3,65 +3,339 @@
...
+        """
+
+        # define limits for logistic growth
+        observed_df = observed_df.rename(columns=self.column_names_map)
+
+        if self.growth == "logistic":
+            self.logistic_growth_floor = observed_df["y"].min() * 0.5
+            observed_df["floor"] = self.logistic_growth_floor
+            self.logistic_growth_cap = observed_df["y"].max() * 1.5
+            observed_df["cap"] = self.logistic_growth_cap
```
bochocki, line 289: Is this used anywhere? These floor/cap scalars seem quite arbitrary, I'd be hesitant to use the scalars as-is.

bochocki: To expand on this a little: logistic growth is convenient for some models because it enables exponential growth at values near the floor and saturating growth at values near the cap. For intermediate regions, the growth rate is roughly linear.

If the distance between the bounds is too large, you lose the dynamic benefits of a logistic curve and most of the growth will be modeled as roughly linear.

jobs/kpi-forecasting/kpi_forecasting/models/prophet_forecast.py
```diff
@@ -165,281 +436,366 @@ def _predict_legacy(self) -> pd.DataFrame:
...
+        project (str): The Big Query project that the data should be written to.
+        dataset (str): The Big Query dataset that the data should be written to.
+        table (str): The Big Query table that the data should be written to.
+        write_disposition (str): In the event that the destination table exists,
+            should the table be overwritten ("WRITE_TRUNCATE") or appended to
+            ("WRITE_APPEND")?
+    """
+    # get legacy tables
+    # TODO: remove this once the forecasting data model is updated:
+    # https://mozilla-hub.atlassian.net/browse/DS-2676
```
bochocki, line 739: I think the code below is _not_ related to DS-2676, is that right?
```suggestion

```

bochocki (commented):
I’ve left some comments —- overall, this looks good. However, I am slightly concerned that the code complexity seems to have increased significantly. This might make it harder to maintain without a dedicated developer. It would be helpful to consider ways to simplify or document the code further for ease of future maintenance.

From the notebook, it seems like the KPI forecasts remain the same before and after these changes. If that’s confirmed, I think we’re in a good position to move forward with merging once the outstanding comments are addressed.

jobs/kpi-forecasting/kpi_forecasting/models/funnel_forecast.py
jaredsnyder: Yeah I think a big improvement would be to refactor to use a standardized parameter optimization library like Optuna or something so we'd get different methods for free

jobs/kpi-forecasting/kpi_forecasting/models/prophet_forecast.py
jaredsnyder: Yeah it is used in the funnel forecast

jobs/kpi-forecasting/kpi_forecasting/models/base_forecast.py
jaredsnyder: For some reason I got into the habit of using "el" in list comprehensions,  changed to `x`

jobs/kpi-forecasting/kpi_forecasting/models/prophet_forecast.py
jaredsnyder: The idea was that its possible to have an empty list for holidays in the config, and in that case things would get screwed up in the `elif` case.  So this ensures that an empty list is treated the same as just leaving it out of the config.

The whole `holidays_raw` and `holidays` thing is annoying.  It comes down to the fact we have an intermediate class to parse the holidays in the config into the dataframe (oh prophet...) that is expected.  And if I want to easily create new prophet classes it's easier to distinguish between the "raw" non-df holiday format and the 'not raw' df format that can be passed to prophet under the `holidays` keyword argument.  It's annoying but I couldn't think of a cleaner way.

bochocki (approved):
Thanks for the quick turnaround! I don't have any good ideas for reducing the code complexity right now, but maybe it's something we can keep in the back of our minds to improve slowly over time.

bochocki:
Darts does look neat! I tried to evaluate it as part of the [KPI model selection exercise](https://colab.research.google.com/drive/1DnFYpClfSzzU3-kz6XzddqWDWFz80QY0#printMode=true&scrollTo=1886c71c-9a9d-4d9b-af03-7d3b523f851b) that we used to decide on prophet, but at the time they didn't have M1 support and that was enough of a blocker for local development that I didn't explore it further.

PR #274, opened 2024-08-26
Support a --date parameter in the fxci metric export

ahal:
Checklist for reviewer:

  responsibly.


ahal:
This will require a corresponding change in `telemetry-airflow` that should land at roughly the same time.

ahal:
@akkomar hi, this is the follow-up to your review comment here:
https://github.com/mozilla/docker-etl/pull/253#discussion_r1705849802

If you do end up reviewing this, feel free to only look at the second commit.

jobs/fxci-taskcluster-export/fxci_etl/console.py
```diff
@@ -1,9 +1,11 @@
+from datetime import datetime, timedelta
 import sys
 from pathlib import Path
 
 from cleo.application import Application
 from cleo.commands.command import Command
 from cleo.helpers import option
+import pytz
```
hneiva, line 8: sort-of unrelated nit: is there a specific reason we use the third party `pytz` package?
`datetime` has a module called `timezone` ([ref](https://docs.python.org/3/library/datetime.html#timezone-objects)). ie: `datetime.now(timezone.utc)`

ahal: Ah thanks for calling that out.. I guess I was following some old stack overflow suggestions.

ahal: Fixed in a new commit.

hneiva (approved):
LGTM!

akkomar (approved):
LGTM.

BTW I looked at these jobs in Airflow and they seem to be taking ~30 mins to run. If this time increases or we'd like to increase scheduling frequency, we might want to look into writing to GCS and loading from there to speed up the process.

PR #278, opened 2024-09-10
Detect and correct wraparound for negative PPA conversion counts

cbguder:
Differential privacy in [Prio][1] can result in negative conversion counts, which then wrap around to large numbers as all operations are performed modulo a large prime number.

This change hard-codes the prime number used for the Prio3SumVec data type (currently the only data type we use for DAP reports) to detect and correct this behavior, such that we can generate human-friendly reports for advertisers.

Namely, given the modulus P is a 128-bit prime, any value N in the histogram that is greater than 2^127 is replaced with (N-P).

JIRA: [AE-553][2]

[1]: https://github.com/divviup/libprio-rs/
[2]: https://mozilla-hub.atlassian.net/browse/AE-553

Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/dap-collector-ppa-dev/dap_collector_ppa_dev/main.py
+++ b/jobs/dap-collector-ppa-dev/dap_collector_ppa_dev/main.py
@@ -10,6 +10,18 @@ import requests
 LEADER = "https://dap-09-3.api.divviup.org"
 CMD = f"./collect --task-id {{task_id}} --leader {LEADER} --vdaf {{vdaf}} {{vdaf_args}} --authorization-bearer-token {{auth_token}} --batch-interval-start {{timestamp}} --batch-interval-duration {{duration}} --hpke-config {{hpke_config}} --hpke-private-key {{hpke_private_key}}"
 MINUTES_IN_DAY = 1440
+
+# The modulo prime for the field for Prio3SumVec, and its size in bits. We use these to detect and counteract negative
+# conversion counts (as a result of differential privacy noise being added) wrapping around.
+#
+# Note that these values are specific to the data type we use for our tasks. If we start using a different type (e.g.
+# Prio3Histogram), the values will need to be adjusted.
+#
+# https://github.com/divviup/libprio-rs/blob/a85d271ddee087f13dfd847a7170786f35abd0b9/src/vdaf/prio3.rs#L88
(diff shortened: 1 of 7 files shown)


jobs/dap-collector-ppa-dev/dap_collector_ppa_dev/main.py
```diff
@@ -154,6 +165,23 @@ async def collect_once(task, timestamp, duration, hpke_private_key, auth_token):
     return res
 
 
+def parse_histogram(histogram_str):
```
dmueller, line 168: The method name is confusing given the comment that we're using vectors and not histograms currently. Does this name make sense given the response shape of the collect call?

jobs/dap-collector-ppa-dev/dap_collector_ppa_dev/main.py
```diff
@@ -154,6 +165,23 @@ async def collect_once(task, timestamp, duration, hpke_private_key, auth_token):
...
+    ]
+
+
+def correct_wraparound(num):
+    cutoff = 2 ** (FIELD_SIZE - 1)
+
+    if num > cutoff:
+        return num - FIELD_PRIME
+
+    return num
```
dmueller, line 182: We could potentially use a field from the tasks.json to decide how we performed the wraparound, looks like we have a "metric_type" field there with "vector" as the value. If we did that, then we could log an error result in the BQ table if we started using a different metric type that we didn't know how to correct.

Also don't feel strongly for this, since I haven't heard any peep about using a different metric type & as is will set us up for running the test campaign

dmueller (approved):
I know this is still in draft, but this all looks reasonable. Also really agree with you that this should be done on their side more and more

jobs/dap-collector-ppa-dev/dap_collector_ppa_dev/main.py
cbguder: Good call, renamed it to `parse_vector`.

jobs/dap-collector-ppa-dev/dap_collector_ppa_dev/main.py
cbguder: This is a good call, but let's do it in a future PR. The rest of the code already assumes we're gonna get a vector (and even assumed we would get a vector of size 5, which this PR now changes).

IMO it would be better to do this holistically (i.e. have none of the task details hard-coded) than to do it for a single method.

PR #292, opened 2024-10-29
Add ETP bugs and their meta bug dependencies to webcompat_kb ETL

ksy36:
Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/webcompat-kb/webcompat_kb/main.py
+++ b/jobs/webcompat-kb/webcompat_kb/main.py
@@ -342,9 +342,8 @@ class BugzillaToBigQuery:
         etp_reports: BugsById,
         etp_dependencies: BugsById,
     ) -> BugsById:
-        """Unify blocked and depends_on for each ETP bug as their dependencies are inconsistent
-
-        and store only ETP meta bugs"""
+        """Unify blocked and depends_on for each ETP bug as their dependencies are inconsistent,
+        keep only ETP meta bugs and store them only in depends_on field."""
 
         filtered = {}
 
@@ -366,6 +365,7 @@ class BugzillaToBigQuery:
             ]
 
             bug["depends_on"] = blocks + depends
+            bug["blocks"] = []
 
             filtered[bug_id] = bug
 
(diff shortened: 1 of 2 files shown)


jobs/webcompat-kb/webcompat_kb/main.py
```diff
@@ -1195,9 +1285,20 @@ def run(self) -> None:
...
+        if etp_reports:
+            etp_reports_unified = self.unify_etp_dependencies(
+                etp_reports, etp_dependencies
+            )
+            etp_data, _ = self.process_relations(
+                etp_reports_unified, ETP_RELATION_CONFIG
+            )
+
+            etp_rels = self.build_relations(etp_data, ETP_RELATION_CONFIG)
+            self.update_relations(etp_rels, ETP_RELATION_CONFIG)
```
jgraham, line 1297: Move this down under `self.update_relations(rels, RELATIONS_CONFIG)`, even though this will require a second check to ensure that `etp_rels` is populated (maybe just set it to an empty dictionary and then we don't need the check I think).

That will reduce the clash with a patch I have to add a `--no-write` argument to prevent trying to actually write data for testing.

jobs/webcompat-kb/webcompat_kb/main.py
```diff
@@ -609,7 +692,7 @@ def update_kb_ids(self, ids):
         table = self.client.get_table(kb_bugs_table)
         logging.info(f"Loaded {table.num_rows} rows into {table}")
 
-    def update_relations(self, relations):
+    def update_relations(self, relations, relation_config):
```
jgraham, line 695: I don't remember why I didn't add type annotations here, but if we could it would be nice :)

jgraham (approved):
Generally LGTM assuming that you've tested it.

Are you able to write some unit tests for the ETP import and in particular for the handling of `depends_on` and `blocks`?

jobs/webcompat-kb/webcompat_kb/main.py
ksy36: Added:)

ksy36:
Added tests and made the rest of the changes, could you take another look, please @jgraham?

jgraham (approved):

PR #298, opened 2024-11-04
Support importing CrUX data into webcompat kb

jgraham:
Checklist for reviewer:

  responsibly.


jobs/webcompat-kb/webcompat_kb/bugzilla.py
```diff
@@ -0,0 +1,1366 @@
...
+        else:
+            logging.info("Skipping writes")
+
+
+def add_arguments(parser: argparse.ArgumentParser) -> None:
+    group = parser.add_argument_group(
+        title="bugzilla", description="Bugzilla import arguments"
+    )
+    parser.add_argument(
+        "--bugzilla-api-key",
```
ksy36, line 1343: Would need to add `dest="bugzilla_api_key"` here too, to match the underscore format in the [airflow config](https://github.com/mozilla/telemetry-airflow/blob/65f2bf0d8cb0cb3a42bc2a1fb0a50cd347e39e41/dags/webcompat_kb.py#L49).

Or we can just change all underscore arguments in airflow to the new format to avoid keeping the legacy ones.

ksy36: Oh actually never mind, I missed that it's reading it from the `BUGZILLA_API_KEY` env variable

jobs/webcompat-kb/webcompat_kb/bugzilla.py
```diff
@@ -0,0 +1,1366 @@
...
+        action="store_false",
+        default=True,
+        help="Don't read or update bug history",
+    )
+
+
+def main(client: bigquery.Client, args: argparse.Namespace) -> None:
+    bz_bq = BugzillaToBigQuery(
+        client,
+        args.bq_project_id,
```
ksy36, line 1359: I think we don't need to pass `bq_project_id` here as it's set in the client and not used in the class?

ksy36 (approved):
It works if I run individual jobs, like `bugzilla` or `crux`. But shows an error when trying to run both (i.e. if I don't specify the argument):

```
webcompat-etl --bq_project_id=... --bq_dataset_id=...
```

`webcompat-etl: error: argument jobs: invalid choice: [] (choose from 'bugzilla', 'crux')`

Not sure if that's expected? Other than that, it looks good to me :)

jgraham:
> Not sure if that's expected

No, that was me not understanding the argparse API correctly :) I added some custom logic to get the behaviour I wanted (no arguments runs all jobs, but if jobs are supplied they have to be one of the provided ones).

ksy36:
It works as expected now, thanks for working on this :)

@scholtzan wonder if could you merge this PR as well, please?

PR #299, opened 2024-11-05
Improve argument validation for webcompat jobs

jgraham:
This also included a refactor so that jobs are defined by classes rather than requiring specific functions in modules.

Checklist for reviewer:

  responsibly.


jobs/webcompat-kb/webcompat_kb/crux.py
```diff
@@ -46,7 +49,7 @@ def get_imported_datasets(client: bigquery.Client, config: Config) -> int:
 SELECT
   yyyymm
 FROM
-  `{config.bq_project}.{config.bq_crux_dataset}.import_runs`
+  `{config.bq_crux_dataset}.import_runs`
```
denschub, line 52: Can you explain your motivation for removing the project prefix here? I *think* it should still work, as the `bigquery.Client` is initialized with the `project` argument, so it should always default to the right project, but I find it rather confusing that we prefix the project name in some queries, but not all.

denschub: Okay - I should start looking at commit messages. This seems to be a deliberate change, given the commit message literally says it's deliberate. I still find it confusing, but I find a lot of this confusing since I'm not yet fully familiar with this codebase, so let's just ignore my comment. :)

denschub (approved):

jobs/webcompat-kb/webcompat_kb/crux.py
jgraham: For posterity: the original motivation here was to reduce the number of places were doing string substitution, given that we're already providing the project id to the client.

PR #300, opened 2024-11-11
Import CrUX data into hosts_min_ranks table

jgraham:
This is a direct view of the specific data we want: global minimum rank, local minimum rank, and sightline minimum rank. The requirement for a subquery to select the latest CrUX import makes it hard to write as a materialized view, but since we only need to update it when the CrUX data itself updates, adding it to the ETL makes sense.

Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/webcompat-kb/webcompat_kb/crux.py
+++ b/jobs/webcompat-kb/webcompat_kb/crux.py
@@ -116,6 +116,47 @@ WHERE
         logging.info(f"Sightline data has {result.total_rows} rows")
 
 
+def update_min_rank_data(client: bigquery.Client, config: Config, date: int) -> None:
+    if config.write:
+        insert_str = f"INSERT `{config.bq_project}.{config.bq_crux_dataset}.host_min_ranks` (yyyymm, host, global_rank, local_rank, sightline_rank)"
+    else:
+        insert_str = ""
+
+    query = f"""
+{insert_str}
+SELECT
+  {date} as yyyymm,
+  NET.HOST(origin) AS host,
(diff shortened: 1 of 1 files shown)


jgraham:
CC @denschub

jobs/webcompat-kb/webcompat_kb/crux.py
```diff
@@ -116,6 +116,47 @@ def update_sightline_data(client: bigquery.Client, config: Config, date: int) ->
...
+        insert_str = ""
+
+    query = f"""
+{insert_str}
+SELECT
+  {date} as yyyymm,
+  NET.HOST(origin) AS host,
+  MIN(
+  IF
+    (origin_ranks.country_code = "global", origin_ranks.rank, NULL)) AS global_rank,
```
denschub, line 132: Can we put these all in one line, as in

```
MIN(IF(origin_ranks.country_code = "global", origin_ranks.rank, NULL)) AS global_rank,
```
?

That would make this a bit easier to understand

denschub (approved):
I have a minor style nit, but no hard feelings either way.

jobs/webcompat-kb/webcompat_kb/crux.py
jgraham: I also kind of hate it, but that's what the Big Query formatter produces, so I'm tempted to leave it.

denschub: If an autoformatter did that, I'll concede my point.

PR #303, opened 2024-12-10
Remove possibility of changepoint after historical data

m-d-bowerman:
Error in the Airflow job called out in [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1936329) is caused by the `changepoint_range` param allowing Prophet to set a changepoint at the last point in the historical data, which causes an error in training the model. This removes that possibility.


jaredsnyder (approved):
Confirmed this will solve the bug we're seeing, lgtm

PR #306, opened 2024-12-12
Consider all related bugs to site reports to be core platform bugs.

jgraham:
Currently we only fetch related bugs if they are explicitly marked as `webcompat:platform-bug`. But this has led to many bugs being missed from the dataset, so swtich to taking all dependencies of site reports as platform bugs.

Checklist for reviewer:

  responsibly.


denschub (approved):
This makes sense to me.

PR #308, opened 2024-12-18
Looker Cleanup and Maintenance Utils

scholtzan:
This adds some Looker cleanup and maintenance CLI tooling which will be scheduled to run in Airflow. The first job is to delete branches that haven't been updated in 180 days.
Looker does not automatically remove branches, even if they were deleted on Github, so this job will remove them from Looker. This should speed up the experience when users create new branches in development mode and have to wait for the branch dropdown to populate.

Checklist for reviewer:

  responsibly.


kwindau (approved):

jobs/looker-utils/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM python:3.8
+MAINTAINER REPLACE ME <replaceme@mozilla.com>
```
sean-rose, line 2: Should this be you?

jobs/looker-utils/looker_utils/main.py
```diff
@@ -0,0 +1,92 @@
...
+@click.option(
+    "--project",
+    help="Looker project name",
+    multiple=True,
+    default=["spoke-default", "looker-hub"],
+)
+@click.option(
+    "--n_days",
+    "--n-days",
+    help="Delete branches that haven't been updated within the last n days",
```
sean-rose, line 50: It'd be nice if this had a more intuitive name.  Maybe `inactive_days`?

jobs/looker-utils/looker_utils/main.py
```diff
@@ -0,0 +1,92 @@
...
+                print(
+                    f"Deleting branch {branch.name} {lookml_project}, last commit on {commit_date}"
+                )
+                sdk.delete_git_branch(
+                    project_id=lookml_project, branch_name=branch.name
+                )
+
+
+if __name__ == "__main__":
+    cli(obj={})
```
sean-rose, line 92: What's `obj`?  (I don't see that as a declared argument for `cli()`)

jobs/looker-utils/looker_utils/main.py
```diff
@@ -0,0 +1,92 @@
...
+        for branch in branches:
+            commit_date = datetime.fromtimestamp(branch.commit_at, timezone.utc)
+
+            if (
+                commit_date < date_cutoff
+                and not branch.name.startswith("dev")
+                and branch.name not in exclude
+            ):
+                print(
+                    f"Deleting branch {branch.name} {lookml_project}, last commit on {commit_date}"
```
sean-rose, line 84: Since multiple projects are being processed it'd be good to mention the project either in these messages, or in separate messages when the project starts being processed.

sean-rose, line 84:
```suggestion
                    f"Deleting branch {branch.name} in {lookml_project}, last commit on {commit_date}"
```

jobs/looker-utils/looker_utils/main.py
scholtzan: It's to provide an empty object that gets referenced within `cli()` as `ctx.obj["SDK"]`

jobs/looker-utils/looker_utils/main.py
```diff
@@ -0,0 +1,92 @@
...
+                commit_date < date_cutoff
+                and not branch.name.startswith("dev")
+                and branch.name not in exclude
+            ):
+                print(
+                    f"{branch.name} in {lookml_project}, last commit on {commit_date}"
+                )
+                # sdk.delete_git_branch(
+                #     project_id=lookml_project, branch_name=branch.name
+                # )
```
sean-rose, line 88: I'm guessing you had this commented out for testing?
```suggestion
                sdk.delete_git_branch(
                    project_id=lookml_project, branch_name=branch.name
                )
```

jobs/looker-utils/looker_utils/main.py
sean-rose: Could `obj` be added to the context within `cli()` rather than having to be passed in?  Since it's just an internal implementation detail to pass the SDK around it seems weird to be passing it in externally.

scholtzan: Adding `obj` to the context within `cli()` doesn't work with something like `TypeError: 'NoneType' object does not support item assignment` being raised.
This also seems to be the intended way looking at the click docs: https://click.palletsprojects.com/en/stable/commands/#nested-handling-and-contexts

sean-rose: Interesting.  Those same docs seem to indicate that if you call `ctx.ensure_object(dict)` in `cli()` then maybe there's no need to pass it in (since we're just passing in an empty dict anyway).

jobs/looker-utils/looker_utils/main.py
```diff
@@ -0,0 +1,92 @@
...
+    required=True,
+)
+@click.option(
+    "--instance_uri", "--instance-uri", envvar="LOOKER_INSTANCE_URI", required=True
+)
+@click.pass_context
+def cli(ctx: dict, client_id: str, client_secret: str, instance_uri: str):
+    sdk = setup_sdk(client_id, client_secret, instance_uri)
+    ctx.obj["SDK"] = sdk
+    pass
```
sean-rose, line 35:
```suggestion
```

jobs/looker-utils/looker_utils/main.py
scholtzan:
```suggestion
    cli()
```

jobs/looker-utils/looker_utils/main.py
scholtzan, line 35:
```suggestion
    ctx.ensure_object(dict)
    ctx.obj["SDK"] = sdk
```

PR #309, opened 2025-01-02
Switch to dev mode before deleting Looker branches

scholtzan:
The Airflow task is currently failing as branches can only be deleted in dev mode https://workflow.telemetry.mozilla.org/dags/looker/grid?dag_run_id=manual__2025-01-02T17%3A23%3A41.422064%2B00%3A00&task_id=delete_outdated_branches&tab=logs

Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/looker-utils/looker_utils/main.py
+++ b/jobs/looker-utils/looker_utils/main.py
@@ -4,6 +4,7 @@ from datetime import datetime, timedelta, timezone
 import click
 import looker_sdk
 from looker_sdk import methods40 as methods
+from looker_sdk.sdk.api40 import models
 
 
 def setup_sdk(client_id, client_secret, instance) -> methods.Looker40SDK:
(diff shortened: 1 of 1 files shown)


jobs/looker-utils/looker_utils/main.py
```diff
@@ -69,6 +69,9 @@ def delete_branches(ctx, project, inactive_days, exclude):
         days=inactive_days
     )
 
+    # switch to dev mode
+    sdk.update_session({"workspace_id":"dev"})
```
BenWu, line 73: The [source for this method](https://github.com/looker-open-source/sdk-codegen/blob/5223398b5e1cf960bc8cc6130906e133b08e990b/python/tests/integration/test_methods.py#L277) uses a `WriteApiSession` object instead of a dict so I think that would be more correct but maybe that's just a wrapper around a dict and this works as well

BenWu (approved):

PR #317, opened 2025-01-27
Use uv to runs tests if it's available

jgraham:
This significantly increases the performance of setting up the test environment.

Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/webcompat-kb/.gitignore
+++ b/jobs/webcompat-kb/.gitignore
@@ -4,3 +4,4 @@ __pycache__/
 build/
 venv/
 webcompat_kb.egg-info/
+uv.lock
--- a/jobs/webcompat-kb/test.sh
+++ b/jobs/webcompat-kb/test.sh
@@ -1,15 +1,23 @@
-#! /usr/bin/bash
+#!/usr/bin/env bash
 
 set -ex
 
-if [ ! -d "_venv" ]; then
-  python3 -m _venv
+if [ -x "$(command -v uv)" ]; then
+    uv sync --extra=test
+    uv run mypy -m webcompat_kb
+    uv run pytest --ruff --ruff-format .
+else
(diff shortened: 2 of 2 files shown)


denschub (changes requested):
This is cool, I didn't even know about `uv`. But if you're touching that file anyway, can you replace the shebang with

```
#!/usr/bin/env bash
```

please? That would make it work under macOS, where the bash is just at `/bin/bash`.

We should also add `uv.lock` to the `.gitignore`, unless your plan is to check it in.

denschub:
Interestingly, and I only noticed now because I relied on CI in the past, there is actually a formatting error:

```
webcompat_kb/metric.py FF                                  [100%]

============================ FAILURES ============================
__________________________ test session __________________________
webcompat_kb/metric.py:2:8: F401 [*] `json` imported but unused
  |
1 | import argparse
2 | import json
  |        ^^^^ F401
3 | import logging
4 | from datetime import date
  |
  = help: Remove unused import: `json`


__________________________ test session __________________________
File would be reformatted
---------------------- Captured stdout call ----------------------
Would reformat: webcompat_kb/metric.py
==================== short test summary info =====================
FAILED webcompat_kb/metric.py::ruff
FAILED webcompat_kb/metric.py::ruff::format
================== 2 failed, 46 passed in 0.40s ==================
```

I'm not entirely sure why the whole `webcompat_kb/` folder isn't even format-checked in CI, but that might be a bug (although not a bug of this PR, of course).

jgraham:
> This is cool, I didn't even know about `uv`.

I only started using it a few days ago, but it's just immediately solved several of the most painful issues with Python environments, so it was a bit of a revelation.

> Interestingly, and I only noticed now because I relied on CI in the past, there is actually a formatting error

I noticed that too; I need to investigate the CI jobs.

denschub (approved):

PR #327, opened 2025-02-10
Store task tags in a separate table

ahal:
The motivation for this is to make it easier to write queries / create dashboards that filter by more than one tag at a time. Having them as a nested column makes this difficult!

Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/fxci-taskcluster-export/fxci_etl/schemas.py
+++ b/jobs/fxci-taskcluster-export/fxci_etl/schemas.py
@@ -82,9 +82,18 @@ class Runs(Record):
 
 
 @dataclass
-class Tag:
-    key: BigQueryTypes.STRING
-    value: BigQueryTypes.STRING
+class Tags:
+    created_for_user: Optional[BigQueryTypes.STRING]
+    kind: Optional[BigQueryTypes.STRING]
+    label: Optional[BigQueryTypes.STRING]
+    os: Optional[BigQueryTypes.STRING]
+    owned_by: Optional[BigQueryTypes.STRING]
+    project: Optional[BigQueryTypes.STRING]
+    trust_domain: Optional[BigQueryTypes.STRING]
+    worker_implementation: Optional[BigQueryTypes.STRING]
+    # No longer used. Kept for backwards compatibility with older records.
+    key: Optional[BigQueryTypes.STRING]
+    value: Optional[BigQueryTypes.STRING]
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


jobs/fxci-taskcluster-export/fxci_etl/schemas.py
```diff
@@ -92,12 +93,29 @@ class Tasks(Record):
...
+    # No longer used. Kept for backwards compatibility with older records.
+    tags: Optional[list[Tag]]
 
     def __str__(self):
         return self.task_id
 
 
+@dataclass
+class Tags(Record):
+    task_id: BigQueryTypes.STRING
```
bhearsum, line 105: It's unclear to me if this is new or not, but I'm curious why `task_id` is a tag. What sort of query would you write where it would be helpful to filter on this, or otherwise use it? I guess maybe it helps to group multiple runs from the same task together?

bhearsum (approved):
Sorry for the delay in reviewing. This looks fine!

jobs/fxci-taskcluster-export/fxci_etl/schemas.py
ahal: Good question, each dataclass represents a different table in BigQuery, so the task_id is just needed to join it with the correct record in the `Tasks` table. I did consider putting all the tags into directly into the `Tasks` table (so there wouldn't be a separate `Tags` table.

Hmm, initially I didn't want to do that because I didn't want to pollute the `Tasks` table with a bunch of arbitrary tags.. but I guess they could still go in a nested column? I think I need to play around with this a bit first

bhearsum: I'm fine with the current approach, to be clear, especially now that I understand that these are just different tables. This seems like classic 1:M relationship management in a database?

ahal: Yeah, there's certainly nothing wrong with this approach.. But if I could leave this as a nested column, but still with the nested keys being well defined, I think that would make it simpler to query and would make the logic here a bit cleaner.

I think I want to at least test it out first before deciding which one to go with..

ahal: Thanks for this comment, I think the latest push will make things much nicer! I'm not sure if you want to re-review or not.

bhearsum (approved):
The new version looks good to me too!

ahal:
Ugh, I forgot to that I needed to push this to a branch on the main repo to trigger the CI. New PR here:
https://github.com/mozilla/docker-etl/pull/329

(and removing my remote so I don't make the mistake again)

PR #341, opened 2025-04-17
Handle capitalization changes in keywords

jgraham:
It seems sometimes keywords can change in their capitalization, so if we don't find the keyword in the historic set check for entries that differ only in case.

Checklist for reviewer:

  responsibly.


denschub (approved):

jgraham:
@scholtzan can you merge this please? :)

PR #342, opened 2025-04-17
feat(DENG-8090): Add verified domain

kwindau:
Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/extensions/extensions/main.py
+++ b/jobs/extensions/extensions/main.py
@@ -175,6 +175,31 @@ def get_category_from_soup(webpage_soup):
     return category
 
 
+def get_verified_domain(soup):
+    """Input: soup
+    Output: verified_domain string if found, else None"""
+
+    target_text = "Created by the owner of the listed website. The publisher has a good record with no history of violations."
+
+    # Find the tag that contains the exact text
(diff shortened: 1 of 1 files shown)


jobs/extensions/extensions/main.py
```diff
@@ -175,6 +175,31 @@ def get_category_from_soup(webpage_soup):
     return category
 
 
+def get_verified_domain(soup):
+    """Input: soup
```
kik-kik, line 179: Instead of this comment, I'd adding typing hints to the function definition.

jobs/extensions/extensions/main.py
```diff
@@ -175,6 +175,31 @@ def get_category_from_soup(webpage_soup):
...
+    Output: verified_domain string if found, else None"""
```
kik-kik, line 180: Perhaps it would be better if we followed an existing structure for a docstring to be consistent, here's an example:
https://github.com/mozilla/bigquery-etl/blob/b454a3ba01d57a69597fa3602b5a459478c7134e/bigquery_etl/view/broken_views.py#L50-L60

kik-kik (approved):

PR #343, opened 2025-04-17
Refactor bugzilla import

jgraham:
The previous import was getting rather hard to understand, because it
was mixing IO with bug grouping logic in a way that made it difficult
to tell which bugs ended up in which groups.
    
In the new version, we simply load all the bugs that we are interested
in into a single large dict at the start. We then do all the grouping
and filtering we need in Python code, before finally writing all the
data to bigquery.
    
Unlike bugs themselves, bug history is a bit more complicated because we allow
incremental updates. This is currently grouped into a single class
that only covers history updating, but some additional work would be
required to make it match the pure input/processing/output flow that
we have for bugs themselves.
    
In addition this change tries to used defined types wherever possible,
and avoids using untyped dictionaries except at the edges. This makes
it much easier to catch a class of errors using mypy.

Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/webcompat-kb/tests/test_bugzilla.py
+++ b/jobs/webcompat-kb/tests/test_bugzilla.py
@@ -1,30 +1,52 @@
+import tempfile
+from collections import defaultdict
 from datetime import datetime, timezone
 from unittest.mock import Mock, patch
-from typing import Any, Mapping
+from typing import Any, Iterable, Mapping
 
 import pytest
-from bugdantic.bugzilla import History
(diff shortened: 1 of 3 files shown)


jgraham:
https://docs.google.com/spreadsheets/d/1Ni1-Q4P615nFvCNEJAujvinzA3oFZy58A1JjpNDkQoM/edit?gid=285700832#gid=285700832 compares changes to the data before and after this refactor.

denschub (approved):

PR #349, opened 2025-04-25
Create a bigquery wrapper and use it everywhere

jgraham:
This centalises a lot of our common patterns in one place.

Each job gets a wrapper with the dataset approproately configured for the job. The wrapper also knows about the write flag, which it can use to avoid writes when creating tables or loading data, which is the common case, but it isn't able to prevent general queries from performing inserts.

Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/webcompat-kb/tests/test_bugzilla.py
+++ b/jobs/webcompat-kb/tests/test_bugzilla.py
@@ -7,7 +7,7 @@ from typing import Any, Iterable, Mapping
 import pytest
 import bugdantic.bugzilla
 
-from webcompat_kb.base import get_client
+from webcompat_kb.bqhelpers import get_client, BigQuery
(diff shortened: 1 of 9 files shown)


denschub (approved):
This is.. a lot, and it's possible I missed something - but it appears to work, and we can always improve as we go... :)

jgraham:
@scholtzan Sorry, can you also merge this one? I thought I'd already merged the dependencies of the other PR, but apparently not, so the previous merge didn't affect main.

scholtzan:
Looks like there are some merge conflicts. Can you update this branch? I'll merge once they got resolved

jgraham:
@scholtzan Not really merge conflicts, it just had a commit that was on an already merged branch, which it didn't like. Should be fixed now.

PR #351, opened 2025-04-28
Fix bug leading to duplication of metric change entries

jgraham:
The underlying bug was a broken check for the bug being new, which meant we considered old bugs with changes to always be new.

In addition add a verification step that will reject any changes where we already have an entry for the same bug with the same timestamp.

Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/webcompat-kb/webcompat_kb/metric_changes.py
+++ b/jobs/webcompat-kb/webcompat_kb/metric_changes.py
@@ -2,6 +2,7 @@ import argparse
 import logging
 import re
 import uuid
+from collections import defaultdict
 from dataclasses import dataclass
 from datetime import datetime, timezone
 from typing import Iterator, Mapping, Optional, Sequence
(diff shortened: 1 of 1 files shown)


jobs/webcompat-kb/webcompat_kb/metric_changes.py
```diff
@@ -288,6 +319,8 @@ def compute_historic_scores(
     historic_states: Mapping[int, list[BugState]],
     current_scores: Mapping[int, float],
 ) -> Mapping[int, list[float]]:
+    """Compute the webcompat scores correspoding to different states of bugs."""
```
denschub, line 322:
```suggestion
    """Compute the webcompat scores corresponding to different states of bugs."""
```

🙈

denschub (approved):

jgraham:
@scholtzan can you merge this one so it deploys, please

PR #353, opened 2025-04-30
Handle missing dependencies

jgraham:
Sometimes we can't access dependent bugs because they're sec bugs or something else we don't have access to. In this case we need to:

* Not keep trying to fetch the bug forever, even though it looks like a missing bug.
* Ensure that we don't fail later when the bug isn't present

The other approach here would be to use a placeholder entry for the missing bug. That would have the advantage that it would maintain the same invariants we previously had, and reduce data loss, but at the cost of having to create useful fake data for the missing bugs.

Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/webcompat-kb/webcompat_kb/bugzilla.py
+++ b/jobs/webcompat-kb/webcompat_kb/bugzilla.py
@@ -459,16 +459,13 @@ class BugCache(Mapping):
             logging.error(f"Error: {e}")
             raise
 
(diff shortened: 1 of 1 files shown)


jgraham:
Alternate approach is https://github.com/mozilla/docker-etl/pull/354, but I think this one is simpler.

denschub (approved):

jgraham:
@scholtzan can you merge please?

PR #355, opened 2025-04-30
Job for importing mozilla standards-positions data

jgraham:
Checklist for reviewer:

  responsibly.


jgraham:
CC @zcorpan

denschub (approved):
GitHub [recently lowered the rate limits for non-authenticated API access](https://github.blog/changelog/2025-05-08-updated-rate-limits-for-unauthenticated-requests/), and it seems it's [now only 60 requests per hour](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api#primary-rate-limit-for-unauthenticated-users). I'm somewhat concerned that this will quickly cause errors in our ETLs, since all those jobs run from the same IP...

Implementation looks good, though.

PR #382, opened 2025-07-04
deng_8994_add python script to download details about lookml models

Marlene-M-Hirose:
Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/looker-utils/looker_utils/lookml_model_details.py
+++ b/jobs/looker-utils/looker_utils/lookml_model_details.py
@@ -1,0 +1,101 @@
+import json
+import os
+
+import requests
+from datetime import datetime
+from dateutil.relativedelta import relativedelta
+import click
+import csv
+
+CSV_FIELDS = [
+    "submission_date",
+    "explores",
+    "has_content",
+    "label",
+    "name",
+    "project_name",
+]
+
+def get_response(url, headers, params):
+    """GET response function."""
+    response = requests.get(url, headers=headers, params=params)
+    try:
+        response.raise_for_status()
+    except requests.exceptions.HTTPError as err:
+        if response.status_code == 404:
+            raise err
+        return("Not Found. Possible Permissions Error")
+        if response.status_code != 500:
+            raise err
+        return ({"items": []}, "skipped")
+    return (response.json(), "completed")
+
+def write_dict_to_csv(json_data, filename):
+    """Write a dictionary to a csv."""
+    with open(filename, "w") as out_file:
+        dict_writer = csv.DictWriter(out_file, CSV_FIELDS)
+        dict_writer.writeheader()
+        dict_writer.writerows(json_data)
+
+def looker_login_post(client_id, client_secret):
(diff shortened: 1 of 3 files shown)


alekhyamoz:
Read client_id / client_secret once at entry; pass only the token downstream. You’re calling looker_login_post() with the creds once, and using the token after.
But if any other function later (like get_response() or looker_lookml_download()) tries to use client_id or client_secret again, that's a problem. Those functions don’t need to know about authentication credentials. They should work only with the token, to reduce security surface area.

Marlene-M-Hirose:
> Read client_id / client_secret once at entry; pass only the token downstream. You’re calling looker_login_post() with the creds once, and using the token after. Or may be have a looker_login() function. But if any other function later (like get_response() or looker_lookml_download()) tries to use client_id or client_secret again, that's a problem. Those functions don’t need to know about authentication credentials. They should work only with the token, to reduce security surface area.

I'm a little confused here as I do use a login function. Do you mean using click to get the client id and secret?

And I am using what Anna recommended for `client_id` and `client secret`

Marlene-M-Hirose:
> > > Replace print and raw returns with logging for debug/info/error
> > 
> > 
> > Do you mean the CSV at the end?
> 
> This is general comment not limited to CSV

Ok, I get it. Essentially, move the errors into the login function, not where it is currently located

Marlene-M-Hirose:
> The code is passing around the client_id and client_secret to multiple functions, when in most cases only the access token is needed after initial authentication. Anna's suggestion is more around storing and retrieving the secrets and not around how to use them in code

I only see the client_id and client_secret in the click, the main, and the login function. They only get called in the login, afterwards, only the token is being used. Unless I'm completely missing something

alekhyamoz:
> > The code is passing around the client_id and client_secret to multiple functions, when in most cases only the access token is needed after initial authentication. Anna's suggestion is more around storing and retrieving the secrets and not around how to use them in code
> 
> I only see the client_id and client_secret in the click, the main, and the login function. They only get called in the login, afterwards, only the token is being used. Unless I'm completely missing something

This comment is more of future proofing rather than a current bug. You don't have to return the entire response from https://github.com/mozilla/docker-etl/blob/6bb6514b86d760d9b8d03432e502e9283214fec5/jobs/looker-utils/looker_utils/lookml_model_details.py#L44 but just return the `access_token`

alekhyamoz:
> > > > Replace print and raw returns with logging for debug/info/error
> > > 
> > > 
> > > Do you mean the CSV at the end?
> > 
> > 
> > This is general comment not limited to CSV
> 
> Ok, I get it. Essentially, move the errors into the ~login function, not where it is currently located?~ get_response function

I was referring to using `logging` rather than `print`- Example - https://github.com/mozilla/bigquery-etl/blob/00cadb60a349e1b851aee7b07d05da73d3589fa8/bigquery_etl/cli/static.py#L5

scholtzan (approved):

PR #386, opened 2025-08-07
Quicksuggest serp categories

ncloudioj:
We're going to add a new field `serp_categories` to RS and this ingests that.

Checklist for reviewer:

  responsibly.


luc-lisi (commented):
Thanks for this! I added a quick question/comment as I'm not sure if the GBQ load will fail without the table schema also being aware of this field.

luc-lisi (approved):
Looks good!

PR #387, opened 2025-08-18
[AE-782] Build dap collector job for incrementality experiments

mashalifshin:
Problem Statement:

Currently, we are running incrementality experiments on FFHNT, measuring whether our Tiles are effective by seeing if they give a lift to organic visits, or to checkouts, to the tile advertiser. We're doing this in a private, anonymized manner, by sending the experiment results to DAP. However, we have no way to collect those results from DAP in an automated way and bring them into our usual data pipeline, so those results are currently being manually collected by an engineer.

Solution: 

Create a `docker-etl` job that can go out to Nimbus and find out how to collect the results for each experiment branch, go out and collect those results from DAP, and finally write those results into BQ.

Checklist for reviewer:

  responsibly.

### DIFF
--- a/jobs/ads-incrementality-dap-collector/dev_run_docker.sh
+++ b/jobs/ads-incrementality-dap-collector/dev_run_docker.sh
@@ -8,3 +8,4 @@ docker run -it --rm \
         --job_config_bucket ads-nonprod-stage-incrementality-dap-collector-config \
         --auth_token $DAP_AUTH_TOKEN \
         --hpke_private_key  $DAP_PRIVATE_KEY \
+        --process_date $PROCESS_DATE \


jobs/ads-incrementality-dap-collector/example_config.json
```diff
@@ -0,0 +1,15 @@
+{
+  "bq": {
+    "namespace": "ads_dap",
+    "table": "incrementality"
+  },
+  "dap": {
+    "hpke_config": "AQAgAAEAAQAgpdceoGiuWvIiogA8SPCdprkhWMNtLq_y0GSePI7EhXE",
+    "batch_start": "1755291600",
+    "batch_duration": "3600"
```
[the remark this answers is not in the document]
mashalifshin: Done

jobs/ads-incrementality-dap-collector/dev_run_docker.sh
gleonard-m, line 1: Is it normal practice to include dev scripts in the PR?

mashalifshin: Yeah I think it's a good idea, if/when someone comes back to extend or modify this code (our future selves included), we'll want a quick and handy way to know how to spin up the docker container and which params/env vars are needed for it to start up properly. Otherwise you'd have to start digging through the code to figure out how to run it (I've been there, much less fun than running a script).

I think a docker-compose file could be a better way to both document and have a canonical way to spin up the code in dev using a single command, and I'd be happy to set that up. Might be overkill since we only have a single container in this setup. But I also think the script works just fine for this purpose, so it's just a matter of if we want to spend the time to make a docker-compose ... what do you think?

jobs/ads-incrementality-dap-collector/tests/test_helpers.py
```diff
@@ -0,0 +1,71 @@
...
+
+class TestHelpers(TestCase):
+    @patch("requests.get", side_effect=mock_nimbus_success)
+    def test_get_experiment_success(self, mock_fetch):
+        experiment = get_experiment("traffic-impact-study-5", "nimbus_api_url")
+        self.assertEqual("traffic-impact-study-5", experiment.slug, )
+        self.assertEqual(1, mock_fetch.call_count)
+
+    @patch("requests.get", side_effect=mock_nimbus_fail)
+    def test_get_experiment_fal(self, mock_fetch):
```
gleonard-m, line 28: nit: is fal a typo?

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/main.py
```diff
@@ -0,0 +1,59 @@
...
+    required=True,
+)
+@click.option(
+    "--batch_start",
+    type=int,
+    envvar='BATCH_START',
+    help="Start of the collection interval, as the number of seconds since the Unix epoch",
+    required=True,
+)
+def main(gcp_project, job_config_bucket, hpke_token, hpke_private_key, batch_start):
```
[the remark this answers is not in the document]
mashalifshin: Ooh yes, this is a great idea.

I was thinking along similar lines after I put the `batch_duration` alongside the experiment slug in the config. But I was wondering if it's possible to specify `batch_duration` somewhere in the Nimbus metadata? This seems like it's part of the experiment design, and could naturally fit with the experiment set-up step. 

Then we just need to tell the job the experiment slugs in the config file, it can go out to Nimbus to get the `batch_duration` as well as the experiment's start date, job runs daily, and then it can check against the current date as you specified, to decide if it needs to go out and collect or not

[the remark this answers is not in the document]
mashalifshin: One more related idea -- we can also provide the `hpke_config` alongside the experiment (either in the Nimbus experiment metadata, or within the experiment's clause in the `config.json`), and then we can use that to select the private key from airflow secrets. Maybe we name the secret/env var for the private key based on the encoded public key value, as an easy way to associate them... 

That would allow the job to collect from multiple experiments that may use different credentials...

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/main.py
```diff
@@ -0,0 +1,59 @@
...
+            experiment = get_experiment(experiment_config.slug, config.nimbus.api_url)
+
+            tasks_to_collect = prepare_results_rows(experiment)
+
+            collected_tasks = collect_dap_results(tasks_to_collect, config.dap, experiment_config)
+
+            write_results_to_bq(collected_tasks, config.bq)
+
+            write_job_logs_to_bucket(gcp_project, job_config_bucket)
+    except Exception as e:
```
mashalifshin, line 46: With the way this exception handling is structured, if we fail to collect on a particular experiment, we'll break out of the loop that process experiments and end the script. 

I think it could be good to change the structure so that if we fail on one experiment, the job could continue on and try processing the other experiments ...

(Leaving this is a note to myself while I work through the other comments)

mashalifshin: Done.

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/helpers.py
```diff
@@ -0,0 +1,206 @@
...
+            raise Exception(f"Error inserting rows into {table_id}: {insert_res}")
+
+def write_results_to_bq(collected_tasks: dict, config: BQConfig):
+    """Takes the collected results for each experiment branch and writes out rows to BQ. """
+    records = [v for inner in collected_tasks.values() for v in inner.values()]
+    logging.info(f"Inserting results rows into BQ: {records}")
+    bq_client = bigquery.Client(project=config.project)
+    full_table_id = create_bq_table_if_not_exists(config.project, config.namespace, config.table, bq_client)
+    for record in records:
+        row = create_bq_row(start_date=datetime.date.today().isoformat(), end_date=datetime.date.today().isoformat(),
```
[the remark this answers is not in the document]
mashalifshin: I went with `collection_start` and `collection_end` which I think you suggested downthread. That seems most descriptive, although could get wordier with `collect_batch_start` or `collection_batch_start` ...

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/helpers.py
```diff
@@ -0,0 +1,206 @@
...
+    """Inserts the results into BQ. Assumes that they are already in the right format"""
+    if row:
+        insert_res = bqclient.insert_rows_json(table=table_id, json_rows=[row])
+        if len(insert_res) != 0:
+            raise Exception(f"Error inserting rows into {table_id}: {insert_res}")
+
+def write_results_to_bq(collected_tasks: dict, config: BQConfig):
+    """Takes the collected results for each experiment branch and writes out rows to BQ. """
+    records = [v for inner in collected_tasks.values() for v in inner.values()]
+    logging.info(f"Inserting results rows into BQ: {records}")
```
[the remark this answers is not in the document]
mashalifshin: Good call, working on this in the next iteration.

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/main.py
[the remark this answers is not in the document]
mashalifshin: Okay, just pushed up the rough logic for this. Still need to do some more validations to make it more robust, and probably clean up the code a bit. 

This also created some new requirements/assumptions/thoughts that I wanted to list out here. I think they all make sense but I'd love your thoughts. It's a brain dump, maybe we can discuss when we next meet up.

- startDate is now required for experiments and the job will fail on experiments that don't have one 
- the job will try to collect for the latest collection batch interval, which is defined to start the day after a collection batch ends
- it seemed a bit brittle to only try to collect for one day/one job run. If it happens to fail on that day, there's no way currently to try again, so I made a `COLLECT_RETRY_DAYS` constants so the job will try to collect n times. This means we will write multiple rows for the same collection batch interval, but I think that's okay if we treat those rows as more "records of job runs" and take the one with the latest `created_at` timestamp. If we really want to ensure that only one row gets written, we can modify the job to check for an existing row, but this seems not worth the effort and not very BQ-y.
- Also thinking about adding logic to stop trying to collect after the experiment's last batch is finished, using experiment `endDate + batchDuration + 1day`. Maybe we can skip this in the initial version since we can also stop it trying to collect by just removing the completed experiment from the config
- Code is also currently not handling `batchDuration` that is less than a day, since the logic compares based on dates, not datetimes. Is it okay to require `batchDuration` to be at least a day? If not, this logic needs adjusting to handle that

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/models.py
```diff
@@ -0,0 +1,211 @@
...
+    """Encapsulates the experiments that should be collected from DAP and how far back to collect,
+        in seconds
+
+        Attributes:
+            slug:               Experiment slug
+            batch_duration:     Duration of the collection batch interval, in seconds
+    """
+
+    slug: str
+    batch_duration: int
```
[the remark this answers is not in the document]
mashalifshin: Done! You can now leave batch duration out of the experiment config and it will default to 7 days.

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/helpers.py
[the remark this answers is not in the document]
mashalifshin: Done :-)

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/helpers.py
```diff
@@ -0,0 +1,331 @@
...
+    return tasks_to_collect
+
+
+# TODO Trigger Airflow errors
+def correct_wraparound(num: int) -> int:
+    field_prime = 340282366920938462946865773367900766209
+    field_size = 128
+    cutoff = 2 ** (field_size - 1)
+    if num > cutoff:
+        logging.info(f"Corrected {num} to {num - field_prime} ")
```
gleonard-m, line 201: We should drop this log since it will output the DAP aggregated result.

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/main.py
```diff
@@ -0,0 +1,77 @@
...
+
+                write_results_to_bq(collected_tasks, config.bq)
+            except Exception as e:
+                logging.error(
+                    f"Collector job failed for {experiment_config.slug}. Error: {e}\n{traceback.format_exc()}"
+                )
+        write_job_logs_to_bucket(gcp_project, job_config_bucket)
+    except Exception as e:
+        logging.error(f"{e}\n{traceback.format_exc()}")
+        write_job_logs_to_bucket(gcp_project, job_config_bucket)
```
gleonard-m, line 65: Move the `write_job_logs_to_bucket` outside the exception handling so we get the log when job is successful

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/models.py
```diff
@@ -0,0 +1,266 @@
...
+        self.task_veclen = visitCountingExperimentListItem.get("task_veclen")
+        # This will be populated when we successfully fetch the count from DAP
+        self.value_count = None
+
+    def __str__(self):
+        return str(
+            f"IncrementalityBranchResultsRow(advertiser='{self.advertiser}', branch='{self.branch}', "
+            f"bucket='{self.bucket}', batch_start='{self.batch_start}', batch_end='{self.batch_end}', "
+            f"country_codes='{self.country_codes}', experiment_slug='{self.experiment_slug}', metric='{self.metric}', "
+            f"task_id='{self.task_id}', task_veclen='redacted', value_count='redacted')"
```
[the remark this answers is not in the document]
gleonard-m: Actually scratch that, `task_id` is available elsewhere, same with `task_veclen` it does not need to be redacted.

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/main.py
```diff
@@ -0,0 +1,77 @@
...
+                tasks_to_collect = prepare_results_rows(experiment)
+
+                collected_tasks = collect_dap_results(
+                    tasks_to_collect, config.dap, experiment_config
+                )
+
+                write_results_to_bq(collected_tasks, config.bq)
+            except Exception as e:
+                logging.error(
+                    f"Collector job failed for {experiment_config.slug}. Error: {e}\n{traceback.format_exc()}"
```
mashalifshin, line 60: Change exception flow so that we throw from this job to airflow if any exception happens (and don't try to keep looping over the experiments).

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/main.py
mashalifshin: Good call, done.

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/models.py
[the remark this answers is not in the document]
mashalifshin: oh right, both are available on the experiment.

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/helpers.py
mashalifshin: Good eye, thank you!

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/main.py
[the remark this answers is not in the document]
mashalifshin: Talking in person with @gleonard-m -- Let's enhance this error handling so we can log the experiment slug and the task ids that might have been involved.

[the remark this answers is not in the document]
mashalifshin: We decided this is okay as-is since the info from the DAP collector error should bubble up here and give us sufficient debug info for now.

jobs/ads-incrementality-dap-collector/README.md
```diff
@@ -0,0 +1,142 @@
...
+Separately, a [DAP](https://docs.divviup.org/) task is configured to store the metrics for each experiment branch in a
+different DAP bucket.
+
+Firefox is instrumented with [DAP telemetry functionality](https://github.com/mozilla-firefox/firefox/tree/main/toolkit/components/telemetry/dap), which allows it to send metrics and reports into the correct DAP buckets as configured in the experiment.
+
+Then this job can go out and collect metrics from DAP (using bucket info from the experiment's data), and write them
+to BQ.
+
+Some examples of existing metrics are
+- "url visit counting", which increments counters in DAP when a firefox client visits an ad landing page.
```
gleonard-m, line 18: An example of the type of metric is not needed here.

jobs/ads-incrementality-dap-collector/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+Great care is taken to preserve privacy and anonymity of these metrics. The DAP technology itself agreggates counts
+in separate systems and adds noise. The DAP telemetry feature will only submit a count to DAP once per week per client.
+All DAP reports are deleted after 2 weeks.
```
gleonard-m, line 22: Details of DAP usage are not needed either since they are not relevant to this ETL job.

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/helpers.py
```diff
@@ -0,0 +1,330 @@
...
+            raise Exception(
+                f"Failed to parse collected DAP results: {collected}"
+            ) from e
+        logging.info(f"Prepared final result rows: {tasks_to_collect[task_id]}")
+        logging.info(f"Finished collecting DAP task: {task_id}")
+    logging.info("Finished DAP collection for all tasks.")
+    return tasks_to_collect
+
+
+# TODO Trigger Airflow errors
```
gleonard-m, line 195: Is this TODO still relevant?

jobs/ads-incrementality-dap-collector/dev_runbook.md
gleonard-m, line 1: This fill should be removed and the contents moved to an internal page

jobs/ads-incrementality-dap-collector/tests/test_mocks.py
```diff
@@ -0,0 +1,351 @@
...
+def mock_dap_subprocess_fail(
+    args: list[str], capture_output: bool, text: bool, check: bool, timeout: int
+) -> CompletedProcess:
+    return CompletedProcess(
+        args=[
+            "./collect",
+            "--task-id",
+            "mubArkO3So8Co1X98CBo62-lSCM4tB-NZPOUGJ83N1o",
+            "--leader",
+            "https://dap-09-3.api.divviup.org",
```
gleonard-m, line 200: use a dummy endpoint

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/helpers.py
mashalifshin: No, this is good now. If we `except` from this function, it'll bubble up to that top level exception handling in `main.py`

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/constants.py
```diff
@@ -0,0 +1,72 @@
...
+        name="value",
+        field_type="RECORD",
+        mode="REQUIRED",
+        fields=[
+            bigquery.SchemaField("count", "INT64", mode="NULLABLE"),
+            bigquery.SchemaField("histogram", "JSON", mode="NULLABLE"),
+        ],
+    ),
+    bigquery.SchemaField(
+        "created_at",
```
gleonard-m, line 67: Update to `created_timestamp` ?

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/main.py
```diff
@@ -0,0 +1,73 @@
...
+    prepare_results_rows,
+    collect_dap_results,
+    write_job_logs_to_bucket,
+    write_results_to_bq,
+)
+
+
+@click.command()
+@click.option("--gcp_project", help="GCP project id", required=True)
+@click.option(
```
mashalifshin, line 18: Additional comment from @gleonard-m: 
Config file and BQ results table will be in different projects.

This `gcp_project` should just be used for the location of the config bucket. Rename to be more specific.

And we need to be able to separately specify the `gcp_project` for the BQ table where the results will be written. That will go in the BQ part of the config json file.

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/models.py
```diff
@@ -0,0 +1,227 @@
...
+
+    def collect_today(self) -> bool:
+        return (
+            self.latest_collectible_batch_end()
+            < self.todays_date()
+            < (
+                self.latest_collectible_batch_end()
+                + timedelta(days=self.COLLECT_RETRY_DAYS)
+            )
+        )
```
gleonard-m, line 121: Can you add tests for these methods?  They are important in determining if the experiment should be collected.

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/models.py
```diff
@@ -0,0 +1,227 @@
...
+
+    startDate: date
+    endDate: Optional[date]
+    enrollmentEndDate: Optional[date]
+
+    # Experiment results are removed from DAP after 2 weeks, so our window to collect results for
+    # an experiment is up to 2 weeks after the experiment's latest_collectible_batch_end.
+    # However, we don't need to collect every day of those two weeks (this job runs daily). So this
+    # constant defines how many days after latest_collectible_batch_end to go out and collect and write to BQ.
+    COLLECT_RETRY_DAYS = 7
```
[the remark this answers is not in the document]
mashalifshin: We don't need this retry functionality here, airflow can do this for us.

More detail from @gleonard-m: 

>  In airflow, if the task fails, it can be cleared and will re-run using the same date when the error occurred, even if the re-run occurs a days later.  Once you update the logic for the bug you mentioned above, either you or i can update the code to accept a date parameter.  This might simplify some of the tests wrt to date manipulation.

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/models.py
mashalifshin: Good call. I should have done this much sooner and it would have made it easier/faster to get the logic right. It was waaay trickier than I realized at first. 🏛️ A classic blunder 🏛️ 

So, there is now a test file for `test_batch_interval_logic` which I think covers all the different cases that are possible here. Would love your eyes/thoughts on this if you can see any code paths left un-exercised.

jobs/ads-incrementality-dap-collector/ads_incrementality_dap_collector/models.py
```diff
@@ -0,0 +1,227 @@
...
+        # After the loop, we have the batch start date for the batch that includes today.
+        # We need to return the previous batch, which is now complete and ready for collection.
+        return latest_collectible_batch_start - timedelta(seconds=self.batchDuration) - timedelta(seconds=self.batchDuration)
+
+    def latest_collectible_batch_end(self) -> date:
+        return self.latest_collectible_batch_start() + timedelta(
+            seconds=self.batchDuration, days=-1
+        )
+
+    def next_collect_date(self) -> date:
```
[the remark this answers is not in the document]
mashalifshin: I ended up removing it entirely since it was only used for info logging. Lmk if that works for you.

gleonard-m (approved):

PR #398, opened 2025-09-18
Unbreak webcompat job by temporary removing interop import

ksy36:
Since we don't have required parameters in the airflow config yet, the job is failing. Removing the import to temporary unbreak it.

Checklist for reviewer:

- [ ] Commits should reference a bug or github issue, if relevant (if a bug is
  referenced, the pull request should include the bug number in the title)
- [ ] Scan the PR and verify that no changes (particularly to
  `.circleci/config.yml`) will cause environment variables (particularly
  credentials) to be exposed in test logs
- [ ] Ensure the container image will be using permissions granted to
  [telemetry-airflow](https://github.com/mozilla/telemetry-airflow/)
  responsibly.

**Note for deployments:** In order to push images built by this PR, the user who merges the PR
must be in the [telemetry Github team](https://github.com/orgs/mozilla/teams/telemetry).
This is because deploys depend on the
[data-eng-airflow-gcr CircleCI context](https://app.circleci.com/settings/organization/github/mozilla/contexts/e1876f84-dfea-47ce-b950-a9eb9e0d4d64).
See [DENG-8850](https://mozilla-hub.atlassian.net/browse/DENG-8850) for additional discussion.

PR #399, opened 2025-09-19
Revert "Unbreak webcompat job by temporary removing interop import"

jgraham:
Reverts mozilla/docker-etl#398

This can land if we also merge https://github.com/mozilla/telemetry-airflow/pull/2255 right after.


ksy36:
@scholtzan could you merge this, please?

PR #527, opened 2026-05-20
Bug 2041074 - Batch FXCI BigQuery inserts by size

jwmossmoz:
Follow-up to #526.

The `fxci_pulse_export` run at 2026-05-20 18:30 UTC still failed with a 413 while inserting a batch of 100 `task_definitions_v1` rows. The malformed hg-push task definitions observed today serialize to about 9.44 MB each, so any batch with more than one of those rows can exceed BigQuery insertAll's 10 MB request limit.

This keeps the existing row-count chunking, but also batches by serialized row size with a 9 MB target. Rows larger than the target are inserted alone so they still have a chance to fit under BigQuery's hard request limit.

Validation:
- `uv run --python 3.12 --with-requirements requirements/test.txt --with-editable . pytest`
- `git diff --check`

Bug: 2041074

### DIFF
--- a/jobs/fxci-taskcluster-export/fxci_etl/loaders/bigquery.py
+++ b/jobs/fxci-taskcluster-export/fxci_etl/loaders/bigquery.py
@@ -1,7 +1,6 @@
 import base64
 import json
 from dataclasses import asdict
-from itertools import batched
 from pprint import pprint
 from textwrap import dedent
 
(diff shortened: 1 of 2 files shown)


jobs/fxci-taskcluster-export/fxci_etl/loaders/bigquery.py
```diff
@@ -19,12 +18,57 @@
 from fxci_etl.config import Config
 
 
+MAX_INSERT_REQUEST_BYTES = 9_000_000
```
ahal: nit: please move this to the class alongside DEFAULT_CHUNK_SIZE

jobs/fxci-taskcluster-export/fxci_etl/loaders/bigquery.py
```diff
@@ -19,12 +18,57 @@
...
+
 class BigQueryLoader:
     DEFAULT_CHUNK_SIZE = 5000
 
-    def __init__(self, config: Config, table_type: str, chunk_size: int = DEFAULT_CHUNK_SIZE):
+    def __init__(
+        self,
+        config: Config,
+        table_type: str,
+        chunk_size: int = DEFAULT_CHUNK_SIZE,
```
ahal, line 33: Maybe we should change the chunk size back to 1000 for task definitions as it shouldn't be necessary anymore

jobs/fxci-taskcluster-export/fxci_etl/loaders/bigquery.py
```diff
@@ -19,12 +18,57 @@
...
 
 
+MAX_INSERT_REQUEST_BYTES = 9_000_000
+
+
+def _json_size(obj: object) -> int:
+    return len(json.dumps(obj, separators=(",", ":")).encode("utf-8"))
+
+
+def _insert_batches(
```
ahal: nit: let's call this `_chunk_batches`

I would also just move it into the class and use `self.chunk_size` and `self.max_insert_bytes` instead of passing them in

jobs/fxci-taskcluster-export/fxci_etl/loaders/bigquery.py
```diff
@@ -19,12 +18,57 @@
...
+    batches = []
+    batch = []
+    batch_size = 0
+
+    for record in records:
+        record_size = _json_size(record)
+        if batch and (
+            len(batch) >= max_count or batch_size + record_size > max_bytes
+        ):
+            batches.append(batch)
```
ahal: Instead of building a list of batches, you could simply `yield batch` here (and at the end). No need to hold all batches in memory at once.

ahal (changes requested):

jobs/fxci-taskcluster-export/fxci_etl/loaders/bigquery.py
ahal: On second thought, let's do that in a follow-up just to be extra safe.. I want this to work again first and foremost :)

ahal (approved):

jwmossmoz (commented):
LGTM!

scholtzan (approved):

PR #535, opened 2026-06-04
[SPONS-162] Add suggestion uuids to quicksuggest2bq job, and log don't error on new kinto suggestion fields

mashalifshin:
[https://mozilla-hub.atlassian.net/browse/SPONS-162](https://mozilla-hub.atlassian.net/browse/SPONS-162)

**Summary**
This change prepares the quicksuggest2bq job for the new suggestion uuid field. (see [PR in MARS](https://github.com/mozilla-services/mars/pull/1210) for more details)

It also proposes a change to log new fields instead of erroring on them, I would be curious if reviewers welcome that change in behavior, or would prefer to leave it noisier with an error. Happy to revert this proposed change and just leave uuids.

**Other affected jobs?**
As far as I can tell this is the only job that currently consumes the kinto suggestions, are there other jobs I should be aware that may require updates for this?

**Merge and rollout timing?** 
This PR is not ready for immediate merge, we should merge it after changes are live in MARS to include new uuids in remote settings payload. Ideally this PR can be merged after new uuids are present in remote settings, but before this job runs. 

If that's not possible, we can consider a more cautious rollout without the need for specific timing, depending on what we decide above wrt to "log don't error" behavior. For example, we can merge just the "log dont error behavior" here, then roll out new uuids field, then merge the "consume new uuids" change here.

Checklist for reviewer:

### DIFF
--- a/jobs/quicksuggest2bq/quicksuggest2bq/main.py
+++ b/jobs/quicksuggest2bq/quicksuggest2bq/main.py
@@ -39,6 +39,7 @@ class KintoSuggestion:
     iab_category: str
     icon: str
     id: int
+    uuid: str
     keywords: List[str]
     title: str
     url: str
@@ -157,6 +158,7 @@ def store_suggestions(
             bigquery.SchemaField("icon", "STRING"),
             bigquery.SchemaField("impression_url", "STRING"),
             bigquery.SchemaField("id", "INTEGER"),
+            bigquery.SchemaField("uuid", "STRING"),
             bigquery.SchemaField("keywords", "STRING", "REPEATED"),
             bigquery.SchemaField(
                 "full_keywords",
--- a/jobs/quicksuggest2bq/quicksuggest2bq/main.py
+++ b/jobs/quicksuggest2bq/quicksuggest2bq/main.py
@@ -8,7 +8,7 @@ import kinto_http
 import logging
 import requests
 
-from dataclasses import asdict, dataclass
+from dataclasses import asdict, dataclass, fields
 from google.api_core.exceptions import BadRequest
 from google.cloud import bigquery
 from typing import Any, Dict, Iterator, List, Optional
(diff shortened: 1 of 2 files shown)


luc-lisi (approved):
I have no issues with this approach and generally agree field additions should be non-breaking! So happy to give it an approval 🎉 

We may still want to wait for DE to confirm this is alright with them though if we are worried about other stronger opinions. It does seem like the comment previously called out that this was made to be intentionally breaking, and so I worry a bit about these warnings being sent silently into the void of the log store. But seeing as we are the only ones who could actually ever make additive changes to the JSON, I think just making this part of the clearly documented process for adding a new field should be alright.

scholtzan (approved):
From a DE perspective I'm ok with this.

To give some possible alternatives we apply to some of our datasets:
* you could add a string or JSON field to the table where all the unknown fields and values go into. We have some tables that have an `additional_properties` field for exactly this purpose. We even have some monitoring set up on these fields to indicate to us that there are additional fields that may need to be added to the schema
* for some of our tables we just allow the addition of new fields. BigQuery tables have a `ALLOW_FIELD_ADDITION` schema update option

mashalifshin:
> From a DE perspective I'm ok with this.
> 
> To give some possible alternatives we apply to some of our datasets:
> 
>     * you could add a string or JSON field to the table where all the unknown fields and values go into. We have some tables that have an `additional_properties` field for exactly this purpose. We even have some monitoring set up on these fields to indicate to us that there are additional fields that may need to be added to the schema
> 
>     * for some of our tables we just allow the addition of new fields. BigQuery tables have a `ALLOW_FIELD_ADDITION` schema update option

Thank you for reviewing and merging, and for sharing some additional approaches here, super helpful as ever @scholtzan! These are both great to know.

I took a look at the `bigquery-etl` repo for some examples of `additional_properties`, love this approach for storing arbitrary, potentially differently-shaped data that is more metadata-ish than first-order columns. I think I see some of the monitoring in bigquery-etl//Users/mlifshin/Code/bigquery-etl/sql/moz-fx-data-shared-prod/monitoring_derived/, very cool!

`ALLOW_FIELD_ADDITION` is also great to know, thank you. Looks like in this case we truncate the table and create new schema each time here (not sure why its set up this way), for the case of evolving schema over time this would be suuuper helpful.

PR #538, opened 2026-06-10
[DENG-11197] Migrate orphaning_dashboard job from spark

BenWu:
https://mozilla-hub.atlassian.net/browse/DENG-11197

Migration of the spark job in https://github.com/mozilla/telemetry-airflow/blob/main/jobs/update_orphaning_dashboard_etl.py. This pushes the computations into bigquery so spark isn't needed.  The output is the same as the existing job and testable locally by running the frontend with:
```sh
# run date should line up with what's in https://telemetry.mozilla.org/update-orphaning/
python -m update_orphaning_dashboard.main --run-date 2026-05-31 --dry-run
python serve_frontend.py
```

Switching to glean was not part of this.

Checklist for reviewer:


jobs/update-orphaning-dashboard/README.md
```diff
@@ -0,0 +1,217 @@
...
+This job was written for Python 3.14. Compatibility with other versions isn't guaranteed.
+
+```bash
+pip install -r requirements.txt
+pip install -e .
+# --dry-run writes the JSON to ./test_output/ instead of uploading to GCS:
+python -m update_orphaning_dashboard.main --run-date 2026-06-08 --dry-run
+```
+
+Authenticate with `gcloud auth application-default login`. `--dry-run` writes
```
antonioborondo, line 115: I guess we have to install the Google CLI to use this command: https://cloud.google.com/cli

@BenWu Could you please confirm?

BenWu: Yes some instructions on setup and accessing bigquery can be found here https://docs.telemetry.mozilla.org/cookbooks/bigquery/access.html#bigquery-access-request. You should already have the necessary access by default

jobs/update-orphaning-dashboard/README.md
```diff
@@ -99,6 +99,8 @@ transport needed.
 
 ## Running locally
 
+This job was written for Python 3.14. Compatibility with other versions isn't guaranteed.
```
antonioborondo, line 106: Why this version in particular? Is there any incompatibility with any of the required libraries?

I was able to run the script without any problems using Python 3.12.10.

antonioborondo (approved):
I was able to run the script locally and compare the data generated locally with the data generated on the website, and they match.

jobs/update-orphaning-dashboard/README.md
BenWu: I went with the latest version in the docker image so it has the longest time before it reaches end-of-life.  The job probably works with much older versions and it should be fine to use whatever you have installed but I wrote that just in case there are issues. Generally, I would recommend using pyenv and using the same version as the image

akkomar (approved):
🚀

PR #541, opened 2026-06-19
Hackbot

jgraham:
Checklist for reviewer:


jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -0,0 +1,645 @@
...
+    for updater in updaters:
+        updater.fetch_data()
+        for task in tasks:
+            task.populate_updates(updater)
+        updater.run()
+
+    insert_new_runs(bq_client, scheduled_table, new_runs)
+    insert_complete_runs(bq_client, completed_table, complete_runs.values())
+    if new_runs or complete_runs:
+        record_update(bq_client, run_info_table, run_infos)
```
ksy36: I think it's possible that we'll have etl runs that don't have `new_runs` scheduled,  but only `complete_runs`, so `run_infos` would be recorded as an empty list. 

Since the next docker etl run selects the latest agent run, it won't have `repro` key, so `created_since` would be empty and it falls back to selecting all bugs created in the past week.

```
      repro_run = latest_runs.get("repro")
      created_since = repro_run.last_time if repro_run else None
```
I think this might be ok, since those bugs will have the `autowebcompat:processed` flag set and won't be selected for processing again? Just wanted to double check if this is intended or I read it wrong :)

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -0,0 +1,645 @@
...
+                bug,
+                BugUpdate(bug=bugzilla.BugUpdate(ids=[bug.id])),
+            )
+
+    def run(self) -> None:
+        for _, bug_update in self.bug_updates.values():
+            if bug_update.has_updates():
+                self.client.update_bugs(bug_update.bug)
+            for attachment in bug_update.add_attachments:
+                self.client.create_attachment(attachment)
```
ksy36, line 406: would it need to pass bug_id here?

ksy36 (approved):

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
jgraham: Ah, yeah, that's a good point, we need to make sure that we always write the "repro" time even if it didn't change, or change the SQL to get the last value for each key even if it's not in the latest row.

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
jgraham: Good cactch. I've added it to the `AttachmentCreate` object, and the client can read it from there.

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -0,0 +1,785 @@
...
+        scheduled = task_runner.create_new()
+        new_runs.update(scheduled)
+
+        for updater in updaters:
+            task_runner.configure_updater(updater)
+
+    for updater in updaters:
+        updater.fetch_data()
+        for task_runner in task_runners:
+            if task_runner.scheduled:
```
ksy36: this might need `or task_runner.completed_runs`, right? So updates are also populated for completed runs

jobs/webcompat-kb/data/sql/autowebcompat/tables/hackbot_scheduled/table.toml
```diff
@@ -0,0 +1,31 @@
...
+mode = "REQUIRED"
+
+[agent]
+type = "STRING"
+mode = "REQUIRED"
+
+[task_name]
+type = "STRING"
+mode = "REQUIRED"
+
```
ksy36, line 12: This is missing `source_time`

[source_time]
type = "DATETIME"
mode = "REQUIRED"

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -0,0 +1,785 @@
...
+            CompleteRun.from_rundoc(item).model_dump(mode="json")
+            for item in complete_runs
+        ]
+        self.bq_client.insert_rows(self.completed_table, rows)
+
+    def record_update(self, run_infos: Iterable[RunInfo]) -> None:
+        self.bq_client.insert_query(
+            self.run_info_table,
+            columns=[item.name for item in self.run_info_table.fields],
+            query="SELECT CURRENT_TIMESTAMP() as run_at, @run_infos AS run_info",
```
ksy36:
```suggestion
            query="SELECT CURRENT_DATETIME() as run_at, @run_infos AS run_info",
```

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -0,0 +1,785 @@
...
+        rows = list(
+            self.bq_client.query(f"""
+    SELECT agent, task_name, source_key, MAX(source_time)
+    FROM `{self.run_info_table}`
+    JOIN UNNEST(run_info) AS run_info
+    GROUP BY agent, task_name, source_key
+    """)
+        )
+        if rows:
+            for item in rows[0].run_info:
```
ksy36: This is failing with `ERROR:root:no row field 'run_info'` as the result is now aggregated.
Added `MAX(source_time) AS source_time` and flattened `for item in rows` locally, seems to work:
```
    SELECT agent, task_name, source_key, MAX(source_time) AS source_time
    FROM `{self.run_info_table}`
    JOIN UNNEST(run_info) AS run_info
    GROUP BY agent, task_name, source_key
    """)
        for item in rows:
            rv[(item.agent, item.task_name)][item.source_key] = item.source_time
        return rv
```

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -0,0 +1,785 @@
...
+                        url="https://bugzilla.mozilla.org/attachment.cgi?id={attachment.id}",
+                    )
+                    for attachment in result.attachments
+                    if attachment.file_name is not None
+                ]
+
+        return new_bugs
+
+    def insert_new_runs(self, new_runs: Iterable[ScheduledRun]) -> None:
+        rows = [item.model_dump(mode="json") for item in new_runs]
```
ksy36: insert_rows below is returning an error
`ERROR:root:[{'index': 0, 'errors': [{'reason': 'invalid', 'location': 'source_time', 'debugInfo': '', 'message': 'Invalid datetime string "2026-06-24T11:12:13Z"'}]}]
` because of timezone identifier for `source_time`

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -0,0 +1,785 @@
...
+        self.bq_client.insert_rows(self.scheduled_table, rows)
+
+    def insert_complete_runs(self, complete_runs: Iterable[RunDoc]) -> None:
+        rows = [
+            CompleteRun.from_rundoc(item).model_dump(mode="json")
+            for item in complete_runs
+        ]
+        self.bq_client.insert_rows(self.completed_table, rows)
```
ksy36, line 299: There is same error with `created_at` and `completed_at`. Looks like it works with something like `completed_at.replace(tzinfo=None).isoformat()`

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -0,0 +1,785 @@
...
+                    require_whiteboard.append(
+                        "[autowebcompat:repro-failed]"
+                        if not result.reproduced
+                        else "[autowebcompat:repro-success]"
+                    )
+                    if result.reproduced and result.steps:
+                        bug_update.add_attachments.append(
+                            bugzilla.AttachmentCreate(
+                                ids=[bug_info.number],
+                                data=result.steps,
```
ksy36, line 658: It's passing the file without base64 encoding https://bugzilla.mozilla.org/show_bug.cgi?id=2050018#c2
I've tried adding `data=result.steps.encode("utf-8")` to convert it to bytes as  bugdantic expects and it works

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
ksy36: Thanks! So I think it's able to post an attachment, but fails with a validation error:

```
INFO:httpx:HTTP Request: POST https://bugzilla.mozilla.org/rest/bug/2050018/attachment "HTTP/2 201 Created"
ERROR:root:1 validation error for AttachmentCreateResponse
ids
  Field required [type=missing, input_value={'attachments': {'9601148...2026-06-25T04:37:45Z'}}}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.13/v/missing
```

ksy36 (changes requested):
I've tried running it on one bug and left some comments. The furthest I got so far is after it posts an attachment with steps, it fails with validation error for AttachmentCreateResponse and doesn't make records in `hackbot_completed` or `import_runs`. I'll run it on multiple bugs tomorrow

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
jgraham: I think I can pass Python objects into `insert_rows` and it will serialize in a way that BigQuery is happy with, but I'm not totally sure.

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
jgraham: This is annoying. It turns out that pydantic field validators aren't really compatible with type checking. So I;ve added the `from_raw_data` method that always base64 encodes the input and that works (at the cost of some duplication in bugdantic).

jgraham:
The `AttachmentCreateResponse` error should now be fixed; the Bugzilla documentation was just totally wrong about the structure of the result data.

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
ksy36: I think it's now returning `ERROR:root:Object of type UUID is not JSON serializable`.
Maybe I can try something like https://github.com/mozilla/docker-etl/blob/main/jobs/webcompat-kb/webcompat_kb/etl/interop.py#L28

```
  On ScheduledRun:                                                                                                                                                                                                                
  def to_json(self) -> Mapping[str, Json]:                                                                                                                                                                                        
      rv = self.model_dump(mode="json")                                                                                                                                                                                           
      rv["source_time"] = self.source_time.replace(tzinfo=None).isoformat()                                                                                                                                                       
      return rv                                                                                                                                                                                                                   
                                                                                                                                                                                                                                  
  On CompleteRun:                                                                                                                                                                                                                 
  def to_json(self) -> Mapping[str, Json]:                                                                                                                                                                                        
      rv = self.model_dump(mode="json")                                                                                                                                                                                           
      rv["created_at"] = self.created_at.replace(tzinfo=None).isoformat()                                                                                                                                                         
      rv["completed_at"] = self.completed_at.replace(tzinfo=None).isoformat()                                                                                                                                                     
      return rv  
```

ksy36 (approved):

PR #554, opened 2026-07-27
[SPONS-231] Fix quicksuggest2bq job for uuid2suggestion_id rename

mashalifshin:
This job has been broken for a little while we just didn't get notified before: 

https://bugzilla.mozilla.org/show_bug.cgi?id=2047223
Sister ticket in the SPONS project: https://mozilla-hub.atlassian.net/browse/SPONS-231

I'll look into consuming the notifications for this job directly as we continue to iterate in the suggest remote settings payload.

I'll keep an eye on this and make sure it succeeds on next run, and will follow up with more fixes if any additional issues come up.

Checklist for reviewer:

- [ ] Commits should reference a bug or github issue, if relevant (if a bug is referenced, the pull request should include the bug number in the title)

- [ ] Scan the PR and verify that no changes (particularly to `.circleci/config.yml`) will cause environment variables (particularly credentials) to be exposed in test logs

- [ ] Ensure the container image will be using permissions granted to [telemetry-airflow](https://github.com/mozilla/telemetry-airflow/) responsibly.

PR #555, opened 2026-08-03
Bug 2047223 - quicksuggest2bq: tolerate suggestions without a suggestion_id

akkomar:
The job has failed daily since 2026-06-11. The uuid -> suggestion_id rename was handled in #554, but three Remote Settings records in quicksuggest-amp predate the field entirely and are no longer republished: sponsored-suggestions-us-tablet (last modified 2025-05-12) and sponsored-suggestions-pl-{desktop,phone} (2025-05-29). Because suggestion_id was a required dataclass field, those 1,273 of 33,182 suggestions raised "TypeError: __init__() missing 1 required positional argument: 'suggestion_id'" and failed the run.

Make suggestion_id optional so those rows load with a NULL value, which the BigQuery schema already allows. Log one warning per attachment that is missing the field, so a future upstream rename stays visible in the logs without emitting a line per suggestion.

The dead records are the underlying problem and should be dropped from the collection by the suggest team; this only keeps the job from failing in the meantime.

Checklist for reviewer:

### DIFF
--- a/jobs/quicksuggest2bq/quicksuggest2bq/main.py
+++ b/jobs/quicksuggest2bq/quicksuggest2bq/main.py
@@ -37,11 +37,16 @@ class KintoSuggestion:
     iab_category: str
     icon: str
     id: int
-    suggestion_id: str
     keywords: List[str]
     title: str
     url: str
     full_keywords: List[FullKeyword]
+
+    # `suggestion_id` is missing from Remote Settings records that predate it and
+    # are no longer being republished (e.g. `sponsored-suggestions-us-tablet`,
+    # last modified in May 2025). Those suggestions are ingested with a NULL
+    # `suggestion_id` rather than failing the job.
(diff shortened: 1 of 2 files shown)


mashalifshin:
Hi Arkadiusz, if it's not urgent and we're okay with this failing for another day or two, I'd like to keep `suggestion_id` required in the job in favor of [this PR](https://github.com/mozilla-services/mars/pull/1422) (in review currently) that will instead delete the stale records that are causing this issue. 

I see you mention in the PR description that the real problem is the old records, and indeed the records are from over a year old and you're right they shouldn't be in the collection anyway. `suggestion_id` is about to become load-bearing ads interaction reporting telemetry, so very soon we'll want this job to fail loudly for this situation.

akkomar:
Superseded by https://github.com/mozilla-services/mars/pull/1422

PR #557, opened 2026-08-06
Add puppeteer script to bugzilla

ksy36:
This is adding the puppeteer script as an attachment to bugzilla and uses steps as a fallback.

Checklist for reviewer:


ksy36:
I think we need script contents validation, but it should probably be on the hackbot side.

jobs/webcompat-kb/tests/test_autowebcompat.py
```diff
@@ -150,14 +150,27 @@ def fetch_data(self) -> None:
                 )
 
 
-def test_repro_bugzilla_update():
+png_base64 = "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABAQAAAAA3bvkkAAAACklEQVR4AWNgAAAAAgABc3UBGAAAAABJRU5ErkJggg=="
+
+
+def run_repro_update(
+    script_url: Optional[str] = None,
```
jgraham, line 157: I mean OK, but this seems to be introducing an unnecessary special case (the png is added by default, script is added if supplied, nothing else is possible). It would be nice to have a more generic way to pass in the list of artifacts.

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -760,7 +761,35 @@ def populate_updates(self, updater: Updater) -> None:
                         require_user_story["autowebcompat-repro-channels"] = ",".join(
                             repro_channels
                         )
-                    if result.steps:
+                    # Prefer the Puppeteer script over the prose steps, if
```
jgraham, line 764: I vaguely wonder if we should make the STR the comment and the script the attachment, given how little we like the existing comments, but we can think about that as a followup.

jgraham (approved):

PR #558, opened 2026-08-06
Allow override of site rank in webcompat score using user-story data

jgraham:
This is useful for cases where we think the domain rank isn't representative of the real impact of the problem e.g. for code that is deployed on many different domains or subdomains.

Checklist for reviewer:


jobs/webcompat-kb/data/sql/webcompat_knowledge_base/routines/WEBCOMPAT_METRIC_SCORE_SITE_RANK_MODIFIER/routine.sql
```diff
@@ -1,36 +1,69 @@
-CREATE OR REPLACE FUNCTION `{{ ref(name) }}`(url STRING, crux_yyyymm INT64) RETURNS NUMERIC AS (
+CREATE OR REPLACE FUNCTION `{{ ref(name) }}`(url STRING, crux_yyyymm INT64, user_story JSON) RETURNS NUMERIC AS (
   (
```
ksy36, line 2: Looks like metric_changes etl is also using this routine, so it needs user_story passed in on line 384?

ksy36 (approved):

jobs/webcompat-kb/data/sql/webcompat_knowledge_base/routines/WEBCOMPAT_METRIC_SCORE_SITE_RANK_MODIFIER/routine.sql
jgraham: Good catch, fixed.

PR #560, opened 2026-08-07
Fix site rank modifier routine

ksy36:
Checklist for reviewer:

### DIFF
--- a/jobs/webcompat-kb/webcompat_kb/etl/metric_changes.py
+++ b/jobs/webcompat-kb/webcompat_kb/etl/metric_changes.py
@@ -374,6 +374,7 @@ def compute_historic_scores(
     rank_yyyymm = project["webcompat_knowledge_base"][
         "WEBCOMPAT_METRIC_YYYYMM"
     ].routine()
+    extract_array = project["webcompat_knowledge_base"]["EXTRACT_ARRAY"].routine()
     with client.temporary_table(schema, rows) as tmp_table:
         score_query = f"""
 SELECT number,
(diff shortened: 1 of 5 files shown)


jobs/webcompat-kb/webcompat_kb/etl/metric_changes.py
```diff
@@ -374,14 +374,15 @@ def compute_historic_scores(
...
+    extract_array = project["webcompat_knowledge_base"]["EXTRACT_ARRAY"].routine()
     with client.temporary_table(schema, rows) as tmp_table:
         score_query = f"""
 SELECT number,
        index,
        url,
        keywords,
        user_story,
-       CAST(`{score_no_rank}`(keywords, user_story) * `{rank_modifier}`(url, `{rank_yyyymm}`(), user_story) AS NUMERIC) as score
+       CAST(`{score_no_rank}`(keywords, user_story) * `{rank_modifier}`(url, `{rank_yyyymm}`(), `{extract_array}`(user_story, "$.site-rank-override")) AS NUMERIC) as score
```
jgraham, line 385: I'm not sure this will actually work? It seems like it maybe has the same problem as if you put the array extraction as an argument in the score calculation in `scored_site_reports` (which seems to not work, whereas computing the override in a top level query does.

jgraham (approved):
I'm going to merge this since I think it unbreaks scored_site_reports, but I'm not delighted with the result; I think it's possibly still broken, or at least hard to understand when it will work and when it won't.

PR #561, opened 2026-08-11
Add autowebcompat diagnosis scheduling

jrmuizel:
I haven't looked this over very carefully yet but wanted to get a prototype written down to see what it feels like.

### DIFF
--- a/jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
+++ b/jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
@@ -170,6 +170,42 @@ class ReproSummary(BaseSummary):
     findings: WebcompatReproResult
 
 
+# Whiteboard token that requests a diagnosis run. It's removed from the bug as
+# soon as the run is scheduled, so re-adding it requests another run.
+DIAGNOSE_REQUEST_TOKEN = "[autowebcompat:diagnose]"
+
+
+class AutowebcompatDiagnosisRequest(CreateRequest):
+    bug_id: int
+    model: Optional[str] = None
+    max_turns: Optional[int] = None
+    effort: Optional[str] = None
+    agent: str = "autowebcompat-diagnosis"
+
+
+class DiagnosisResult(BaseModel):
+    reproduced: bool
+    failure_reason: Optional[str] = None
+    root_cause: Optional[str] = None
+    evidence: Optional[str] = None
+    testcase_url: Optional[str] = None
+
+
+class WebcompatDiagnosisResult(HackbotAgentResult):
+    result: Optional[DiagnosisResult] = None
+    start_time: Optional[datetime] = None
+    end_time: Optional[datetime] = None
+
+
+class DiagnosisSummary(BaseSummary):
+    findings: WebcompatDiagnosisResult
+
+
+class DiagnosisBugInfo(BaseModel):
+    number: int
+    source_time: datetime
+
+
 class BigQueryService:
     def __init__(
         self, project: Project, bq_client: BigQuery, bz_client: bugzilla.Bugzilla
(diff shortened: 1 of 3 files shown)


jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -320,6 +358,49 @@ def get_new_bugs(
...
+        if not bugs:
+            return {}
+
+        # Only keep bugs Bugzilla actually returns to us and that aren't
+        # moco-confidential; anything else we can't safely hand to the agent.
+        visible = set()
+        for result in self.bz_client.bugs(
+            list(bugs.keys()), include_fields=["id", "groups"]
+        ):
+            if result.groups and "mozilla-employee-confidential" in result.groups:
```
jgraham, line 400: I think you should also re-check for the whiteboard token here since the bq data is only a cache and can be behind the live bug state.

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -850,6 +947,148 @@ def populate_updates(self, updater: Updater) -> None:
...
+            }
+
+            for uuid, (bug_number, output) in run_outputs.items():
+                bug, bug_update = updater.bug_updates[bug_number]
+
+                result = output.findings.result
+                require_whiteboard = []
+                require_user_story = {}
+                if result is None:
+                    require_whiteboard.append("[autowebcompat:diagnosis-failed]")
```
jgraham: Do we want to stop doing the whiteboard stuff now?

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -170,6 +170,44 @@ class ReproSummary(BaseSummary):
     findings: WebcompatReproResult
 
 
+# Whiteboard token that requests a diagnosis run. It's removed from the bug as
+# soon as the run is scheduled, so re-adding it requests another run.
+DIAGNOSE_REQUEST_TOKEN = "[autowebcompat:diagnose]"
```
jgraham, line 175: Claude and I have different opinions on the value of putting this kind of string as global variables :)

jrmuizel: I lean more toward Claude's approach if the value is being used in multiple places like it is in this case. Happy to change it though if you prefer.

jgraham: I personally would put it in `DiagnosisTask` and pass it in to `get_diagnosis_requested_bugs`, since that scopes it to the relevant code, but it's not important either way.

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -170,6 +170,44 @@ class ReproSummary(BaseSummary):
...
+
+
+class AutowebcompatDiagnosisRequest(CreateRequest):
+    # The diagnosis agent reads the bug itself over the Bugzilla MCP server, so
+    # it takes a bug id rather than a serialised snapshot of the bug.
+    bug_id: int
```
jgraham, line 179: This isn't true; you can either provide `bug_data` or `bug_id` (see https://github.com/ksy36/bugbug/blob/18f700877b225889f6946f640b71f2c737a17bcd/services/hackbot-api/app/schemas.py#L106-L107) I prefer providing `bug_data` since that makes it easier to control the inputs and easier to test changes to prompts &c. 

Since we're planning to work more on this anyway I think it's OK to land the first version without that change, but we should at least remove the incorrect comment.

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -320,6 +358,56 @@ def get_new_bugs(
...
+            for row in rows
+        }
+        if not bugs:
+            return {}
+
+        # site_reports is only as fresh as the last Bugzilla import, so confirm
+        # against live data that the token is still there before dispatching:
+        # a previous run may already have consumed it. Also drop bugs Bugzilla
+        # doesn't return, or that are moco-confidential.
+        eligible = set()
```
jgraham: Minor nit, but we could just construct the return value i.e. something more like
```
diagnosis_bugs = {}
for result in self.bz_client.bugs(…):
    if [condition for including the bug]:
        diagnosis_bugs[result.id] = bugs[result.id]

return diagnosis_bugs
```

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -850,6 +954,143 @@ def populate_updates(self, updater: Updater) -> None:
...
+                            logging.warning(
+                                "Failed to get testcase data from %s",
+                                result.testcase_url,
+                            )
+                else:
+                    require_user_story["autowebcompat-diagnosis-status"] = "failed"
+                    require_user_story["autowebcompat-diagnosis-reason"] = (
+                        result.failure_reason
+                        if result.failure_reason
+                        else "no_root_cause"
```
jgraham, line 1082: Unless I'm getting the precedence wrong here I think this is "unknown" and the bottom one is more like "no_repro"?

jgraham (approved):
I think this is probably fine to land. A few comments, but nothing blocking.

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
jrmuizel: agreed

PR #572, opened 2026-08-20
Add autowebcompat-repro-report-os whiteboard flag

ksy36:
Checklist for reviewer:

### DIFF
--- a/jobs/webcompat-kb/tests/test_autowebcompat.py
+++ b/jobs/webcompat-kb/tests/test_autowebcompat.py
@@ -3,7 +3,7 @@ from collections import defaultdict
 from pathlib import Path
 import json
 from datetime import datetime
-from typing import Mapping, Optional, Iterable, cast
+from typing import Literal, Mapping, Optional, Iterable, cast
 from uuid import UUID
 from unittest.mock import Mock, patch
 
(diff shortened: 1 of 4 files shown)


jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -957,19 +966,17 @@ def populate_updates(self, updater: Updater) -> None:
...
                             reason = result.failure_reason
-                        elif result.plan_result is not None and (
-                            "desktop" not in result.plan_result.affects_platforms
-                            or (
-                                result.plan_result.affects_os is not None
-                                and isinstance(result.plan_result.affects_os, list)
-                                and "linux" not in result.plan_result.affects_os
-                            )
-                        ):
-                            reason = "unsupported_platform"
```
ksy36, line 965: Removing this part as unsupported_platform will be split into 3 values (android, ios and desktop).

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -560,6 +560,15 @@ def try_get_file(url: str, allowed_types: Optional[set[str]] = None) -> Optional
     return data
 
 
+def report_os_value(plan_result: Optional[TestPlanResult]) -> Optional[str]:
```
jgraham: I'd rather have a function that does the "get a string or convert a list to a string" part and put the logic specific to that one field in the caller, or alternatively put this on the `TestPlanResult` object (but that feels like mixing serialization logic into the data model).

jobs/webcompat-kb/tests/test_autowebcompat.py
```diff
@@ -250,6 +250,28 @@ def get_file(url: str, allowed_types: Optional[set[str]] = None) -> Optional[byt
...
+) -> autowebcompat.TestPlanResult:
+    return autowebcompat.TestPlanResult(
+        is_webcompat=True,
+        affects_platforms=["desktop"],
+        affects_os=affects_os,
+        affects_channels=["nightly"],
+    )
+
+
+def test_report_os_value() -> None:
```
jgraham: This amount of testing feels a lot like overkill for a function that  basically does list | str -> str.

jobs/webcompat-kb/tests/test_autowebcompat.py
```diff
@@ -250,6 +250,28 @@ def get_file(url: str, allowed_types: Optional[set[str]] = None) -> Optional[byt
     return updates, result
 
 
+def plan_result(
```
jgraham: This is very specific to the one test we have; if we really want the test I'd paramaterize over the input and output data and construct the TestPlan object inside the test.

jgraham (approved):
OK, but it feels like we're adding more complexity than we really need for this change.

jgraham (approved):

PR #574, opened 2026-08-21
ignore duplicate autowebcompat_reproduction jobs

benoit-chauvet:
Ignore duplicate autowebcompat_reproduction jobs (keep only the most recent one for each bug)

Checklist for reviewer:

### DIFF
--- a/jobs/webcompat-kb/data/sql/autowebcompat/views/autowebcompat_reproduction/view.sql
+++ b/jobs/webcompat-kb/data/sql/autowebcompat/views/autowebcompat_reproduction/view.sql
@@ -7,8 +7,11 @@ WITH
   JOIN `{{ ref('hackbot_completed') }}` completed USING (run_id)
   WHERE scheduled.task_name = 'repro'
     AND STARTS_WITH(scheduled.source_key, "bugzilla:")
+  QUALIFY ROW_NUMBER() OVER (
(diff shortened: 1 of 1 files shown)


jgraham:
This feels like maybe is should go in the frontend? I can imagine wanting to use this view to compute something like the distribution of number of reproduction attempts per bug, if it's not always one.

benoit-chauvet:
That makes sense. So I'll apply this rule for the diagnosis view as well

benoit-chauvet:
done

jobs/webcompat-kb/data/sql/autowebcompat/views/autowebcompat_reproduction/view.sql
```diff
@@ -41,4 +40,4 @@ SELECT
 FROM `{{ ref('webcompat_knowledge_base.site_reports') }}` reports
 LEFT JOIN `{{ ref('webcompat_knowledge_base.scored_site_reports') }}` scored USING (number)
 INNER JOIN hackbot_repro USING (number)
-ORDER BY reports.creation_time DESC
+ORDER BY reports.creation_time DESC;
```
jgraham:
```suggestion
ORDER BY reports.creation_time DESC
```

jgraham (approved):

PR #575, opened 2026-08-21
Create a view for AutoWebCompat diagnosis data

benoit-chauvet:
Checklist for reviewer:

### DIFF
--- a/jobs/webcompat-kb/data/sql/autowebcompat/views/autowebcompat_diagnosis/view.sql
+++ b/jobs/webcompat-kb/data/sql/autowebcompat/views/autowebcompat_diagnosis/view.sql
@@ -1,0 +1,55 @@
+WITH
+requesters AS (
+  SELECT distinct number, LEFT(who, STRPOS(who,"@")-1) as who
+  FROM `{{ ref('webcompat_knowledge_base.bugs_history') }}` h,
+  UNNEST (h.changes) as ch
+  WHERE ch.field_name LIKE '%whiteboard%'
+  AND ch.added LIKE'%autowebcompat:diagnose%'
+),
+requesters_list as (
+  SELECT number, ARRAY_TO_STRING(ARRAY_AGG(who ORDER BY who), ", ")  as who
+  FROM requesters
+  GROUP BY number
+),
+hackbot_diagnosis AS (
+  SELECT
+  PARSE_NUMERIC(JSON_VALUE(scheduled.extra_data, "$.bug_id")) as number,
+  DATE(completed.completed_at) as processing_completion_date,
+  DATE(completed.created_at) as processing_start_date,
+  DATETIME_DIFF(completed.completed_at, completed.created_at, SECOND) as execution_time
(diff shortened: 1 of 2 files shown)


jobs/webcompat-kb/data/sql/autowebcompat/views/autowebcompat_diagnosis/view.sql
```diff
@@ -0,0 +1,55 @@
...
+  SELECT
+  PARSE_NUMERIC(JSON_VALUE(scheduled.extra_data, "$.bug_id")) as number,
+  DATE(completed.completed_at) as processing_completion_date,
+  DATE(completed.created_at) as processing_start_date,
+  DATETIME_DIFF(completed.completed_at, completed.created_at, SECOND) as execution_time
+  FROM `{{ ref('hackbot_scheduled') }}` scheduled
+  JOIN `{{ ref('hackbot_completed') }}` completed USING (run_id)
+  WHERE scheduled.task_name = 'diagnosis'
+  AND STARTS_WITH(scheduled.source_key, "bugzilla:")
+  QUALIFY ROW_NUMBER() OVER (
```
jgraham: As with the other query, I think this shouldn't go in the underlying view, since it makes it very specific to one use case.

jobs/webcompat-kb/data/sql/autowebcompat/views/autowebcompat_diagnosis/view.sql
```diff
@@ -0,0 +1,55 @@
...
+  AND ch.added LIKE'%autowebcompat:diagnose%'
+),
+requesters_list as (
+  SELECT number, ARRAY_TO_STRING(ARRAY_AGG(who ORDER BY who), ", ")  as who
+  FROM requesters
+  GROUP BY number
+),
+hackbot_diagnosis AS (
+  SELECT
+  PARSE_NUMERIC(JSON_VALUE(scheduled.extra_data, "$.bug_id")) as number,
```
jgraham: nit, but it would be nice to have the `SELECT` fields indented when we're putting them on a different line (same with conditions in the `WHERE` clause on different lines, and in general when we have a statement that continues over multiple lines).

jobs/webcompat-kb/data/sql/autowebcompat/views/autowebcompat_diagnosis/view.sql
benoit-chauvet: Just done in the latest commit

jgraham (approved):
OK to land with the duplicate exclusion removed (but the formatting fix would still be nice, just not necessary).

jobs/webcompat-kb/data/sql/autowebcompat/views/autowebcompat_diagnosis/view.sql
benoit-chauvet: ok, it's a pitty I just changed it the wrong way in the repro view, so I'll update both :)

jgraham (approved):

PR #576, opened 2026-08-21
Autowebcompat Backlog reproduction, diagnosis flag for all reproduced bugs

ksy36:
Checklist for reviewer:


jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -337,18 +337,96 @@ def get_new_bugs(
...
+    FROM `{self.project["webcompat_knowledge_base"]["scored_site_reports"]}` as bugs
+    WHERE
+      resolution = "" AND
+      whiteboard NOT LIKE "%[autowebcompat:processed]%" AND
+      "webcompat:platform-bug" NOT IN UNNEST(keywords) AND
+      "webcompat:sitepatch-applied" NOT IN UNNEST(keywords) AND
+      CAST(creation_time AS DATETIME) <= IFNULL(
+        @latest_new_bugs_repro_run, DATETIME_SUB(CURRENT_DATETIME(), INTERVAL 1 WEEK)
+      ) AND
+      number NOT IN (
```
jgraham, line 375: We can't assume that the `run_key` is always a bugzilla bug number. If we need that we should check that the `source_key` starts with `bugzilla:` and get the bug number from `extra_data`.

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -870,6 +948,10 @@ def populate_updates(self, updater: Updater) -> None:
                     elif result.reproduced:
                         require_whiteboard.append("[autowebcompat:repro-success]")
                         require_user_story["autowebcompat-repro-status"] = "success"
+                        if not result.chrome_mask_fixed:
```
jgraham, line 971: Maybe we should check in case a diagnosis already ran (that could definitely be the case for things in the backlog)?

jgraham (changes requested):

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
ksy36: sounds good, I've added that

jgraham (approved):

PR #577, opened 2026-08-25
fix autowebcompat diagnosis failure cause cleanup on success

benoit-chauvet:
Fix the following bug:
When autowebcompat diagnosis suceeds after a previous failure, the previous failure reason wasn't removed from the bug's user story

Checklist for reviewer:


jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -624,6 +624,33 @@ def update_user_story(
         bug_update.bug.cf_user_story = new_value
 
 
+def remove_user_story_fields(
```
jgraham: I think I'd prefer the change where `update_user_story` handles this case too, rather than duplicating most of the code. We could either make `require_tokens` a `Mapping[str, str | None]` and have the `None` value indicate that the field must be removed or have a `remove_tokens` argument that's a set. I think the advantage of the former is that you can't have conflicting values for the same field, and it makes the change below into just:

```
require_user_story["autowebcompat-diagnosis-reason"] = None
```

jgraham (changes requested):

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -610,13 +610,18 @@ def update_user_story(
         if key in current_fields:
             current_value = current_fields[key]
             if isinstance(current_value, list):
+                assert value is not None
```
jgraham: I think we could support this; let's just move the `if value is None` check to the start of this `if` block i.e.
```py
if value is None:
    changes.append(userstory.UserStoryChange.delete(key, current_value))
elif isinstance(current_value, list):
    if value not in current_value:
        changes.append(userstory.UserStoryChange.append(key, value))
else:
    changes.append(	
        userstory.UserStoryChange.replace(key, current_value, value)
    )
```

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
benoit-chauvet: updated. Not sure that it's the most elegant way to handle the None value

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -610,13 +610,18 @@ def update_user_story(
...
-                    userstory.UserStoryChange.replace(key, current_value, value)
-                )
+                if value is None:
+                    changes.append(userstory.UserStoryChange.delete(key, current_value))
+                else:
+                    changes.append(
+                        userstory.UserStoryChange.replace(key, current_value, value)
+                    )
         else:
+            assert value is not None
```
jgraham: This assert is wrong, we can just check whether the value is None and append if it isn't. That handles the case where we try to delete an already deleted value.

jgraham (changes requested):

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
benoit-chauvet: I also handle the case of multiple current values:
```py
if isinstance(current_value, str):
    changes.append(userstory.UserStoryChange.delete(key, current_value))
elif isinstance(current_value, list):
     for val in current_value:
           changes.append(userstory.UserStoryChange.delete(key, val))
```

jgraham (approved):

PR #578, opened 2026-08-25
[EXP-7392] Add highwind job for Nimbus experiment analysis

jaredlockhart:
## Description

A proof of concept of the Nimbus Highwind analysis system.  This implementation hard codes a small set of guardrail metrics for Firefox Desktop and produces confidence intervals for all live experiments and runs daily.  This will allow us to validate the operational health of the system while we separately validate the statistical methodology.

## Related Tickets & Documents
* https://mozilla-hub.atlassian.net/browse/EXP-7392

Checklist for reviewer:


jobs/highwind/highwind/metrics.py
```diff
@@ -0,0 +1,257 @@
...
+    ),
+]
+
+# One entry per source table: where to read it, which columns the metrics above need, and any
+# restriction the metric-hub data source carries. Keeping the restriction in the source CTE's WHERE
+# is the same filter in one scan.
+SOURCES = {
+    CLIENTS_DAILY: dict(
+        table="moz-fx-data-shared-prod.telemetry.clients_daily",
+        unit_column="client_id",
```
jaredlockhart: This should be the randomization unit from the recipe, fixing now

jobs/highwind/highwind/metrics.py
```diff
@@ -0,0 +1,257 @@
...
+            "scalar_parent_browser_engagement_total_uri_count_sum",
+            "scalar_parent_browser_engagement_total_uri_count_normal_and_private_mode_sum",
+            "scalar_parent_os_environment_is_taskbar_pinned",
+            "is_default_browser",
+            "pings_aggregated_by_this_row",
+        ],
+    ),
+    SEARCH: dict(
+        table="moz-fx-data-shared-prod.search.search_clients_engines_sources_daily",
+        unit_column="client_id",
```
jaredlockhart: This should be the randomization unit from the recipe, fixing now

jobs/highwind/highwind/metrics.py
```diff
@@ -0,0 +1,257 @@
...
+            "organic",
+            "tagged_sap",
+            "tagged_follow_on",
+        ],
+    ),
+    # metric-hub's firefox_desktop_active_users_view is
+    # (SELECT * FROM ...telemetry.desktop_active_users WHERE is_desktop).
+    ACTIVE_USERS: dict(
+        table="moz-fx-data-shared-prod.telemetry.desktop_active_users",
+        unit_column="client_id",
```
jaredlockhart: This should be the randomization unit from the recipe, fixing now

jobs/highwind/requirements.txt
```diff
@@ -0,0 +1,10 @@
+click==8.3.1
+google-cloud-bigquery==3.40.0
+google-cloud-storage==3.8.0
+numpy==1.24.2
+pandas==2.3.3
+pydantic==2.9.1
+scipy==1.15.3
+
+flake8==7.3.0
```
mikewilli, line 9: no ruff?? (but this is fine)

jobs/highwind/highwind/main.py
```diff
@@ -0,0 +1,135 @@
...
+    "--validate-sql",
+    is_flag=True,
+    help=(
+        "Submit every query to BigQuery for validation without executing it, and write nothing. "
+        "Checks that the generated SQL is well formed and typed, at no scan cost."
+    ),
+)
+@click.option(
+    "--sample-percent",
+    type=float,
```
mikewilli: `sample_id` is an int 0-99 (including from the UDF you're using) so you probably want this to be as well. I didn't see it getting converted before the WHERE clause is constructed, but that would be another option if you prefer the float here. I'd suggest sticking with int since I think that's pretty standard.

jobs/highwind/highwind/main.py
```diff
@@ -0,0 +1,135 @@
...
+    validate_sql,
+    sample_percent,
+    billing_project,
+):
+    """Analyse every live Firefox Desktop experiment for the run date."""
+    outcomes = run_daily_job(
+        as_of=run_date.date(),
+        only_slugs=list(slugs) or None,
+        limit=limit,
+        dry_run=validate_sql,
```
mikewilli: Should this be `dry_run or validate_sql`?

jobs/highwind/highwind/analysis.py
```diff
@@ -0,0 +1,418 @@
...
+from google.cloud import storage  # type: ignore
+from google.cloud import bigquery
+
+from . import discovery, gbstats_compute
+from . import metrics as metric_definitions_module
+from . import output_writing, sql_generation, sql_running
+
+# The covariate window: days before a unit's enrollment used to predict its outcome. Fixed rather
+# than matched to each window's length, so one pre-period serves every window of a metric.
+COVARIATE_DAYS = 28
```
mikewilli, line 33: Jetstream is using a 7-day pre-enrollment window, just FYI. I don't know if that's better, worse, or totally irrelevant due to methodology changes, but thought I'd mention.

jobs/highwind/highwind/analysis.py
```diff
@@ -0,0 +1,418 @@
...
+
+# The project the queries run and bill in. Every table they read is named by its own project, so
+# this selects who pays rather than what is read.
+DEFAULT_BILLING_PROJECT = "mozdata"
```
mikewilli: I wonder if we should use `moz-fx-data-experiments` for tracking purposes. cc @scholtzan

jobs/highwind/highwind/analysis.py
```diff
@@ -0,0 +1,418 @@
...
+    """Analyse every discoverable experiment for one run date.
+
+    The expensive part is a handful of queries for the whole run rather than a handful per
+    experiment: one cohort scan, then one per source per analysis unit, each grouped by slug. Every
+    experiment's date range overlaps almost every other's, so a per-experiment scan would read the
+    same calendar day once per experiment covering it.
+
+    The cost of sharing them is failure isolation on the query step: a source query that fails fails
+    every experiment rather than one. That is a fair description of reality, since a broken source
+    query IS broken for everyone, and it is the case `systemic_failure` exists to catch. Isolation
```
mikewilli, line 60: I know that preliminary testing has it well under 6 hours, but if this starts getting up closer to that limit we'll want to reconsider this assumption. But that's part of what this POC is meant to test!

jobs/highwind/highwind/analysis.py
```diff
@@ -0,0 +1,418 @@
...
+        states=output_writing.state_counts(results),
+        error=None if error is None else f"{type(error).__name__}: {error}"[:200],
+    )
+
+
+def report_selection(experiments, skipped, as_of):
+    """Print which experiments the run will analyse, and why the others were skipped."""
+    print(
+        f"as_of {as_of}: {len(experiments)} experiments to analyse, {len(skipped)} skipped"
+    )
```
mikewilli, line 350: Consider using logging instead of print.

jobs/highwind/highwind/units.py
```diff
@@ -0,0 +1,73 @@
...
+ANALYSIS_UNITS = {
+    ("firefox_desktop", "group_id"): AnalysisUnit(
+        kind=PROFILE_GROUP_ID,
+        enrollment_column="profile_group_id",
+        source_column="profile_group_id",
+        clustered_sample_id=False,
+    ),
+    ("firefox_desktop", "normandy_id"): AnalysisUnit(
+        kind=CLIENT_ID,
+        enrollment_column="legacy_telemetry_client_id",
```
mikewilli, line 67: I think we want to use the Glean `client_id`, no?

jobs/highwind/highwind/units.py
```diff
@@ -0,0 +1,73 @@
...
+PROFILE_GROUP_ID = "profile_group_id"
+
+
+@dataclass(frozen=True)
+class AnalysisUnit:
+    """One analysis unit, and the column each side of the join records it in.
+
+    `enrollment_column` and `source_column` are separate because the enrollment scan and the metric
+    tables name the same unit differently: on desktop a client is `legacy_telemetry_client_id` in
+    `events_stream` and `client_id` in the metric tables. Joining one name against the other matches
```
mikewilli: Unless I'm mistaken, the `client_id` and `legacy_telemetry_client_id` are different -- one is legacy and one is glean. It might make sense to separate them anyway, but I don't think this comment is correct.

jobs/highwind/highwind/units.py
mikewilli: And see my comment above regarding legacy vs glean.

jobs/highwind/highwind/units.py
```diff
@@ -0,0 +1,73 @@
...
+    ("firefox_desktop", "group_id"): AnalysisUnit(
+        kind=PROFILE_GROUP_ID,
+        enrollment_column="profile_group_id",
+        source_column="profile_group_id",
+        clustered_sample_id=False,
+    ),
+    ("firefox_desktop", "normandy_id"): AnalysisUnit(
+        kind=CLIENT_ID,
+        enrollment_column="legacy_telemetry_client_id",
+        source_column="client_id",
```
mikewilli: Should this use the `CLIENT_ID` constant?

jobs/highwind/highwind/units.py
```diff
@@ -0,0 +1,73 @@
...
+rest of the job's scope. Adding an app is a row here rather than a change to the SQL generation.
+
+A pair with no row is refused. A default would pick a grain, produce intervals from it, and give no
+sign the grain was wrong, which is the failure this module exists to prevent.
+"""
+
+from dataclasses import dataclass
+
+CLIENT_ID = "client_id"
+PROFILE_GROUP_ID = "profile_group_id"
```
mikewilli, line 20: I'm not sure what these are used for, but maybe this should be an enum `AnalysisUnit`?

mikewilli: https://github.com/mozilla/metric-hub/blob/main/lib/metric-config-parser/metric_config_parser/__init__.py#L4-L6

jobs/highwind/highwind/analysis.py
```diff
@@ -0,0 +1,418 @@
...
+    run = sql_generation.Run(
+        experiments, windows_by_slug, as_of, COVARIATE_DAYS, sample_percent
+    )
+
+    cells_by_slug, timings, failure = gather_sufficient_statistics(
+        client, run, metrics_by_source, as_of, dry_run
+    )
+
+    # No lock: `as_completed` yields in this thread, so the accumulation below is single-threaded.
+    outcomes, results_by_slug, done = [], {}, 0
```
mikewilli: Not sure what a better name is, but `outcome` is already used in the experiment context and could be confusing.

jobs/highwind/highwind/analysis.py
```diff
@@ -0,0 +1,418 @@
...
+
+    Carries no scan cost: the scan is shared across the run, so bytes and slot-hours are properties
+    of the run and are reported there.
+    """
+    return dict(
+        slug=experiment.slug,
+        cells=len(results),
+        states=output_writing.state_counts(results),
+        error=None if error is None else f"{type(error).__name__}: {error}"[:200],
+    )
```
mikewilli: Should this be an object instead of a function returning a dict?

jobs/highwind/highwind/analysis.py
```diff
@@ -0,0 +1,418 @@
...
+    """Print which experiments the run will analyse, and why the others were skipped."""
+    print(
+        f"as_of {as_of}: {len(experiments)} experiments to analyse, {len(skipped)} skipped"
+    )
+    for slug, why in skipped:
+        print(f"   skipped {slug}: {why}")
+
+
+def report_run(outcomes, timings):
+    """Print the per-run operational summary: what it cost and how many cells failed.
```
mikewilli: I don't think this reports cost, does it?

jobs/highwind/highwind/discovery.py
```diff
@@ -0,0 +1,221 @@
...
+  AND normandy_slug IS NOT NULL
+  AND start_date IS NOT NULL
+  AND start_date <= @as_of
+  -- Live, or ended recently enough that a final run would still be worth producing. The scheduled
+  -- job passes live_only, so in production the tail is unused; it exists so an ad-hoc run can
+  -- reproduce Jetstream's selection rule, which analyses an experiment for a period after it ends.
+  AND (end_date IS NULL OR DATE_ADD(end_date, INTERVAL @tail_days DAY) >= @as_of)
+-- Shortest first, which is what makes `limit` select the cheapest experiments to analyse rather
+-- than an arbitrary set. Duration is the cost proxy: it sets how far back the shared scan reaches.
+ORDER BY DATE_DIFF(@as_of, start_date, DAY) ASC
```
mikewilli, line 62: Shouldn't this be DESC to get the shortest ones first?

jobs/highwind/highwind/discovery.py
```diff
@@ -0,0 +1,221 @@
...
+    def tenure_days(self, as_of):
+        """Calendar days from first enrollment to the last date with complete data.
+
+        The maturity frontier is derived from this: a window can only be reported once some unit has
+        lived through it, and no unit can have lived longer than the experiment has run.
+        """
+        return (as_of - self.start_date).days
+
+
+def discover(client, as_of, tail_days=90, limit=None, only_slugs=None, live_only=False):
```
mikewilli: Why is 90 days the default here for `tail_days`? It seems like you'd want to stop running 1 day after the experiment ends, unless I'm misinterpreting what this does.

jobs/highwind/highwind/discovery.py
```diff
@@ -0,0 +1,221 @@
...
+            bigquery.ScalarQueryParameter("app_name", "STRING", APP_NAME),
+            bigquery.ScalarQueryParameter("as_of", "DATE", as_of),
+            bigquery.ScalarQueryParameter("tail_days", "INT64", tail_days),
+        ]
+    )
+    experiments, skipped = [], []
+    for row in client.query(DISCOVERY_SQL, job_config=job_config).result():
+        if only_slugs and row.slug not in only_slugs:
+            continue
+        # The 90-day tail mirrors Jetstream's selection rule, so a parallel run compares the same
```
mikewilli: Jetstream gets these from the Experimenter API, true, but then it filters out the ones that don't need analysis (because they didn't end yesterday and there were no config changes since then that would require new analysis). I think for this POC it's fine to just look at a 1-day lag.

jobs/highwind/highwind/discovery.py
```diff
@@ -0,0 +1,221 @@
...
+        query_parameters=[
+            bigquery.ScalarQueryParameter("app_name", "STRING", APP_NAME),
+            bigquery.ScalarQueryParameter("as_of", "DATE", as_of),
+            bigquery.ScalarQueryParameter("tail_days", "INT64", tail_days),
+        ]
+    )
+    experiments, skipped = [], []
+    for row in client.query(DISCOVERY_SQL, job_config=job_config).result():
+        if only_slugs and row.slug not in only_slugs:
+            continue
```
mikewilli, line 117: Why do this and the next `continue` blocks not add the slug to the `skipped` list?

jobs/highwind/highwind/gbstats_compute.py
```diff
@@ -0,0 +1,232 @@
...
+def compute_cell(experiment, metric, window, treatment, cells, theta, failure):
+    """One cell: its state, and its interval when it has one."""
+    identity = dict(
+        metric=metric.name,
+        window=window.label,
+        window_kind=window.kind,
+        window_start=window.start,
+        window_end=window.end,
+        branch=treatment,
+        reference_branch=experiment.reference_branch,
```
mikewilli, line 79: Note for future: pairwise branch comparison.

jobs/highwind/highwind/gbstats_compute.py
```diff
@@ -0,0 +1,232 @@
...
+        theta=theta,
+    )
+
+
+def build_test(reference_statistic, treatment_statistic):
+    """Construct the sequential test for one pair of arms.
+
+    Written against the pinned gbstats signature directly rather than detected at import. Sniffing
+    it from a rendered type annotation fails silently if upstream restyles its annotations, and the
+    failure would surface as every cell in the run erroring rather than as one exception at startup.
```
mikewilli: It's late in the day so sorry if I'm being obtuse, but I'm confused what this is telling me. The last bit about the failure mode seems relevant, but why wouldn't we be running against the version of `gbstats` we've imported?

jobs/highwind/highwind/gbstats_compute.py
```diff
@@ -0,0 +1,232 @@
...
+        return dict(point=None, lower=None, upper=None, theta=theta)
+    return dict(
+        point=result.expected * 100,
+        lower=result.ci[0] * 100,
+        upper=result.ci[1] * 100,
+        theta=theta,
+    )
+
+
+def build_test(reference_statistic, treatment_statistic):
```
mikewilli:
```suggestion
def build_t_test(reference_statistic, treatment_statistic):
```

jobs/highwind/highwind/metrics.py
```diff
@@ -0,0 +1,262 @@
...
+    ),
+]
+
+# One entry per source table: where to read it, which columns the metrics above need, and any
+# restriction the metric-hub data source carries. Keeping the restriction in the source CTE's WHERE
+# is the same filter in one scan. The column each table records the analysis unit in comes from the
+# unit rather than from here, since it varies by unit and not by table.
+SOURCES = {
+    CLIENTS_DAILY: dict(
+        table="moz-fx-data-shared-prod.telemetry.clients_daily",
```
mikewilli, line 222: Should this be Glean?

jobs/highwind/highwind/metrics.py
```diff
@@ -0,0 +1,262 @@
...
+# ------------------------------------------------------------------------ definitions ----
+
+
+@dataclass(frozen=True)
+class Metric:
+    """One metric: where it reads from, how it reduces a unit's rows, and its windows."""
+
+    name: str
+    source: str
+    reducer: Reducer
```
mikewilli, line 156: Something to consider for the metric schema is whether "reducer" or "aggregator" (or something else) is more clear to people configuring metrics.

jobs/highwind/highwind/output_writing.py
```diff
@@ -0,0 +1,374 @@
...
+import pathlib
+from dataclasses import dataclass
+
+from google.cloud import bigquery
+
+# Production targets, the defaults an `Outputs` takes when nothing overrides them.
+RESULTS_TABLE = "moz-fx-data-experiments.monitoring.highwind_statistics_v1"
+SUFFICIENT_STATS_TABLE = (
+    "moz-fx-data-experiments.monitoring.highwind_sufficient_stats_v1"
+)
```
mikewilli: Let me create a dataset for you so these can stay separate from the existing production artifacts in `monitoring`. I'll go with `highwind_poc`.

mikewilli:
```suggestion
RESULTS_TABLE = "moz-fx-data-experiments.highwind_poc.highwind_statistics_v1"
SUFFICIENT_STATS_TABLE = (
    "moz-fx-data-experiments.highwind_poc.highwind_sufficient_stats_v1"
)
```

jobs/highwind/highwind/output_writing.py
```diff
@@ -0,0 +1,374 @@
...
+            "Always pre-enrollment regardless of which window the row is for."
+        ),
+    ),
+    bigquery.SchemaField(
+        "pre_sumsq",
+        "FLOAT64",
+        description="Sum of squares of the covariate, which supplies its variance.",
+    ),
+    bigquery.SchemaField(
+        "xp",
```
mikewilli: Unless this is the standard term, may be worth giving this a more descriptive name.

jobs/highwind/highwind/output_writing.py
```diff
@@ -0,0 +1,374 @@
...
+    idempotent without any extra machinery.
+
+    Carries `generated_at` and `pipeline_version` in the header and as object metadata, so the
+    ingest can tell a new run from an old one without downloading the blob.
+    """
+    # When this ran, not the date it analysed: two runs of the same date, a retry or a backfill,
+    # have to be distinguishable, which is the whole purpose of the field.
+    generated_at = datetime.datetime.now(datetime.timezone.utc).isoformat()
+    payload = {
+        "metrics_meta": {
```
mikewilli:
```suggestion
        "metrics_metadata": {
```

jobs/highwind/highwind/output_writing.py
```diff
@@ -0,0 +1,374 @@
...
+    Written per experiment, unlike the tables below, because the blob IS per experiment and its name
+    is the key: rewriting `<slug>.json` replaces the previous run's answer, so this leg is
+    idempotent without any extra machinery.
+
+    Carries `generated_at` and `pipeline_version` in the header and as object metadata, so the
+    ingest can tell a new run from an old one without downloading the blob.
+    """
+    # When this ran, not the date it analysed: two runs of the same date, a retry or a backfill,
+    # have to be distinguishable, which is the whole purpose of the field.
+    generated_at = datetime.datetime.now(datetime.timezone.utc).isoformat()
```
mikewilli, line 265: I learned today that `datetime.UTC` is an alias for `datetime.timezone.utc` and so now I'm sharing this information with you! (you don't need to use it, I just thought it was nice)

jobs/highwind/highwind/output_writing.py
```diff
@@ -0,0 +1,374 @@
...
+        path.write_text(json.dumps(payload, indent=2))
+        return str(path)
+    blob = blob_for(storage, experiment.slug, outputs.blob_prefix)
+    blob.metadata = {
+        "as_of_date": as_of.isoformat(),
+        "generated_at": generated_at,
+        "pipeline_version": PIPELINE_VERSION,
+    }
+    blob.upload_from_string(json.dumps(payload), content_type="application/json")
+    return f"{outputs.blob_prefix}/{experiment.slug}.json"
```
mikewilli: I don't know if this is a GCS requirement, but I think everything previously has written to GCS with underscores instead of hyphens.

jobs/highwind/highwind/sql_generation.py
```diff
@@ -0,0 +1,717 @@
...
+# enrolled, so one whose members enrolled sparsely can stay under the threshold and survive. The
+# spec's version needs its own scan of a client-daily source, where this rides along on the scan the
+# cohort already does.
+MAX_CLIENTS_PER_GROUP = 10
+
+# The id a group's fan-out is counted over. Present on the enrollment event beside the group id, so
+# counting it costs a column rather than a join.
+FLEET_CLIENT_COLUMN = "legacy_telemetry_client_id"
+
+# How many times the smallest arm the largest is allowed to be. Chosen at the knee of the
```
mikewilli: Is "arm" the same as "branch"?

jobs/highwind/highwind/sql_generation.py
```diff
@@ -0,0 +1,717 @@
...
+            v.unit_kind,
+            {enrollment_unit_id(run)} AS unit_id,
+            e.branch_slug,
+            e.enrollment_date{fleet_select(run)}
+          FROM (
+            SELECT
+              JSON_VALUE(event_extra, '$.experiment') AS slug,
+              {ids},
+              JSON_VALUE(event_extra, '$.branch') AS branch_slug,
+              DATE(submission_timestamp) AS enrollment_date
```
mikewilli, line 384: You'll probably want the `MIN` enrollment date in case there are multiple enrollment events for whatever reason.

jobs/highwind/highwind/sql_running.py
```diff
@@ -0,0 +1,103 @@
...
+Records what each query cost, because on a shared slot reservation wall time reflects how many slots
+happened to be free, and slot-hours is the portable number.
+"""
+
+import time
+from concurrent.futures import ThreadPoolExecutor, as_completed
+
+from google.cloud import bigquery
+
+COHORT_DATASET = "mozdata.tmp"
```
mikewilli: Why `tmp`? Consider writing this to the same dataset as the other results: `moz-fx-data-experiments.highwind_poc`

mikewilli (changes requested):
I left a bunch of comments, but only a couple blockers I remember:
- change the output dataset to `highwind_poc`
- double-check if the GCS files need to be underscores instead of hyphens
 
Overall, it's difficult to review an entire application by just reading the code, so I think ultimately we want to get this out there so we can start testing and monitoring. If it was going to production I'd be more concerned, but it's a POC.

Some thoughts:
Is error information sent to BigQuery at all, or just printed to the container logs? It'd be helpful if it's queryable in BigQuery if it isn't already.

I made some notes on Glean vs Legacy because I thought a goal was to be Glean-only, but if the POC is intentionally just a straight reflection of the existing metrics for closer comparison then feel free to ignore them.

You might consider using pyproject.toml instead of setup.py, and ruff instead of flake8, for consistency with other projects, but it's a POC so not a big deal. I left a few other comments that might be more relevant for a production system, so for things like this (pyproject, ruff, logging) take 'em or leave 'em as you see fit. Some might be worth noting for whenever we do translate this into production, even if you don't tackle them now.

It doesn't seem ideal that we
- can't rerun individual experiments, especially when the stats can fail for individual experiments and underlying data might change from run-to-run (e.g., from shredder)
- can't individually sample experiments (or write sampled results to BigQuery)
but I guess we'll see if these are problems or not, and hopefully pretty quickly.

Since this is so similar in broad shape to jetstream, it might have been nice to leverage some of what jetstream (or related tooling) has already implemented instead of reimplementing here. But again, I understand it was probably faster to generate that code for a POC than find the useful parts and reference them. Something to keep in mind for production though.

And on a similar note, there's a bunch of naming/terminology that'd it be nice to keep from Jetstream for the sake of consistency.

Many of the comments are hard to parse, but that can be cleaned up for production code.

jobs/highwind/highwind/analysis.py
scholtzan: Yeah, using `moz-fx-data-experiments` would be good

jobs/highwind/highwind/output_writing.py
scholtzan: GCS supports hyphens. We just had underscores previously for consistency

jobs/highwind/highwind/analysis.py
```diff
@@ -0,0 +1,418 @@
...
+            # Younger than the shortest window, so no unit has completed one and there is nothing
+            # to compute. Its cells are all `not_started` by construction.
+            return experiment_outcome(experiment, []), []
+        results = gbstats_compute.compute_statistics(
+            experiment, all_metrics(metrics_by_source), windows, cells
+        )
+        if not dry_run:
+            output_writing.write_blob(
+                storage_client, experiment, as_of, results, outputs
+            )
```
scholtzan, line 250: Something worth double-checking is that when using `--sample-percent` it will overwrite the GCS blobs here. There is a guard in `run_daily_jobs` that prevents this for BigQuery table results, so something similar may be needed here.

jobs/highwind/highwind/output_writing.py
```diff
@@ -0,0 +1,374 @@
...
+def ensure_table(client, table, schema):
+    """Create `table` if it is not there yet, and leave it alone if it is.
+
+    The job declares its own destinations because nothing else does, and because the load below
+    writes a single partition, which needs a table to write a partition of. Declaring the schema
+    here rather than inferring it also means a column's type and description are properties of the
+    code that fills them.
+    """
+    definition = bigquery.Table(table, schema=schema)
+    definition.time_partitioning = bigquery.TimePartitioning(field=PARTITION_FIELD)
```
scholtzan, line 345: I also recommend clustering on `experiment_slug`. Probably not super necessary for the POC, but as the table grows with more experiments being run, this makes filtering for specific experiment results faster

jobs/highwind/highwind/sql_running.py
```diff
@@ -0,0 +1,103 @@
...
+def materialize_cohort(client, sql, as_of, dry_run=False):
+    """Write the run's cohort to its own table and return where it went.
+
+    One table for every experiment in the run, keyed by slug and carrying the unit each was
+    randomized on. Materialized rather than inlined because every source query needs it, and a CTE
+    would be re-scanned once per query. The assignment scan is a large share of the job's bytes, so
+    paying it once rather than once per source per experiment is the difference between a viable
+    daily run and an unaffordable one.
+    """
+    table = f"{COHORT_DATASET}.highwind_cohort_{as_of.isoformat().replace('-', '_')}"
```
scholtzan: Another thing worth checking is that concurrent or overlapping runs don't interfere when writing the cohorts table. There is a chance that for the same date there are multiple jobs that `WRITE_TRUNCATE` on the same table

jobs/highwind/requirements.txt
jaredlockhart: 🙈  Ruff only!  Yeah I'll do that as a followup 🙏

jobs/highwind/highwind/analysis.py
jaredlockhart: I know, that'll all get sorted out separately in the DS validation, not material to the POC.

jobs/highwind/highwind/analysis.py
jaredlockhart: Yeah totally, local tests have shown ~1hr for a full run, and I think there's even more ways to optimize to get that down further, but yes something to watch for as it runs.

jobs/highwind/highwind/analysis.py
jaredlockhart: Good idea, can do as follow up 🙏

jobs/highwind/highwind/units.py
jaredlockhart: I know, that'll all happen as we migrate the metrics themselves, right now this is all just stand ins using the existing metric definitions.

jobs/highwind/highwind/units.py
jaredlockhart: Yeah we can clean all this up later 👍

jobs/highwind/highwind/analysis.py
jaredlockhart: Oh good point yep

jobs/highwind/highwind/analysis.py
jaredlockhart: I just wanted to get a sense of like slot hours and data scanned and etc for monitoring.

jobs/highwind/highwind/discovery.py
jaredlockhart: No other way around, smallest are the newest.

jobs/highwind/highwind/discovery.py
jaredlockhart: Oh sorry this was from early prototyping and got left in I'll just take it out.

jobs/highwind/highwind/discovery.py
jaredlockhart: No you're right I'm just gonna remove it.

jobs/highwind/highwind/gbstats_compute.py
jaredlockhart: Yep totally, not included here but something we'll need 👍

jobs/highwind/highwind/gbstats_compute.py
jaredlockhart: Sorry confusing, it's becuase gbstats is dumb and the pypi version reports as being 0.8.0 and the docker pulled one is more recent but still reports as 0.8.0 so it's possible to accidentally have the wrong one and then it just breaks even though the 'version' says it's the right version.  Another point that leaked in from prototyping but isn't relevant here, I'll just take the comment out.

jobs/highwind/highwind/metrics.py
jaredlockhart: Ya we'll figure all that out when we start doing the metric porting.

jobs/highwind/highwind/discovery.py
jaredlockhart: Oh no it's to filter them on on purpose.  The other one was for the live check but I just moved that into the query instead.

jobs/highwind/highwind/metrics.py
jaredlockhart: Later when we migrate the metrics yes

jobs/highwind/highwind/output_writing.py
jaredlockhart: Oh yeah good point, it's just the cross product but I'll use a better name.

jobs/highwind/highwind/analysis.py
jaredlockhart: Ya totally, something to worry about later though, right now nothing consumes it this is more just to monitor its operation.

jobs/highwind/highwind/discovery.py
mikewilli: Oh right I saw `start_date` and didn't read the rest of the line 🤦 

That said, you could save probably milliseconds by just sorting on `start_date DESC` instead because it won't need to compute the date diff.

scholtzan (approved):

mikewilli (approved):

PR #579, opened 2026-08-26
Add try catch for bugzilla failures, store scheduled hackbot runs in BQ right away

ksy36:
There has been a [failing job](https://workflow.telemetry.mozilla.org/dags/webcompat_kb/grid?search=webcompat_kb&dag_run_id=scheduled__2026-08-26T01%3A30%3A00%2B00%3A00&task_id=webcompat_kb&tab=logs) caused by a deleted attachment on one of the bugs ([2043199](https://bugzilla.mozilla.org/show_bug.cgi?id=2043199))

```
[2026-08-26, 01:48:23 UTC] {pod_manager.py:477} INFO - [base] ERROR:root:Request failed
[2026-08-26, 01:48:23 UTC] {pod_manager.py:477} INFO - [base] Failed to fetch key 9593084 from network storage: Not Found
[2026-08-26, 01:48:23 UTC] {pod_manager.py:477} INFO - [base] ERROR:root:Client error '400 Bad Request' for url '[https://bugzilla.mozilla.org/rest/bug?id=2030213%2C2021214%2C2043199&limit=100&offset=0&include_fields=id%2Cgroups%2Ccomments%2Cattachments%2Ccf_user_story'](https://bugzilla.mozilla.org/rest/bug?id=2030213%2C2021214%2C2043199&limit=100&offset=0&include_fields=id%2Cgroups%2Ccomments%2Cattachments%2Ccf_user_story%27)
[2026-08-26, 01:48:23 UTC] {pod_manager.py:477} INFO - [base] For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400
```
Looks like it scheduled the same bug inside ReproTask (new bug, not backlog) on each etl run in hackbot over and over because the scheduled run and `[autowebcompat:processed]` flag weren't recorded (as the failure happened in the BacklogReproTask create_new it never reached `insert_new_runs` for the ReproTask).

Checklist for reviewer:

### DIFF
--- a/jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
+++ b/jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
@@ -321,6 +321,14 @@ class BigQueryService:
     WHERE
       resolution = "" AND
       whiteboard NOT LIKE "%[autowebcompat:processed]%" AND
+      number NOT IN (
+        SELECT CAST(JSON_VALUE(extra_data, "$.bug_id") AS INT64)
+        FROM `{self.scheduled_table}`
+        WHERE
+          agent = @agent AND
+          source_key = "bugzilla:creation" AND
+          JSON_VALUE(extra_data, "$.bug_id") IS NOT NULL
+      ) AND
       (
         (@created_since IS NOT NULL AND CAST(creation_time AS DATETIME) > @created_since) OR
         (
@@ -332,9 +340,10 @@ class BigQueryService:
         bug_rows = self.bq_client.query(
             bugs_query,
             parameters=[
+                bigquery.ScalarQueryParameter("agent", "STRING", ReproTask.agent),
                 bigquery.ScalarQueryParameter(
                     "created_since", "DATETIME", created_since
-                )
+                ),
             ],
         )
         return self.add_bugzilla_data(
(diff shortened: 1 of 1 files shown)


jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -795,22 +830,23 @@ def schedule(
...
+                run_key=request.run_key,
+                source_time=request.source_time,
+                requested_at=requested_at,
+                request_data=request.request_data.model_dump(mode="json"),
+                extra_data=request.extra_data.model_dump(mode="json")
+                if request.extra_data
+                else {},
             )
 
+            self.bq_service.insert_new_runs([scheduled_run])
```
jgraham: This insert-per-run seems like quite unnecessary overhead, plus it means that each task has to handle storing its own scheduled runs. If we want this to happen infallibly we should fix the main `run()` function to handle errors but ensure that all runs that were actually scheduled get inserted.

AIUI this isn't actually needed as part of the emergency bugfix here.

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -321,6 +321,14 @@ def get_new_bugs(
     WHERE
       resolution = "" AND
       whiteboard NOT LIKE "%[autowebcompat:processed]%" AND
+      number NOT IN (
```
jgraham, line 324: This means we can never reschedule reproduction for the same bug, at least using this path. I guess that's OK, but we might want to retry if it runs but fails with an error?

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -321,6 +321,14 @@ def get_new_bugs(
...
+        SELECT CAST(JSON_VALUE(extra_data, "$.bug_id") AS INT64)
+        FROM `{self.scheduled_table}`
+        WHERE
+          agent = @agent AND
+          STARTS_WITH(source_key, "bugzilla:") AND
```
jgraham: You should just match the whole source_key here.

jgraham (changes requested):

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
ksy36: true, changed that

jgraham (approved):
Let's land this for now. I think it does make sense to improve the error handling a bit more so that if we do schedule any runs we make sure to do the insert, even if something fails in the intermediate steps.

PR #581, opened 2026-08-27
Fix  "HTTP/2 406 Not Acceptable" error when adding diagnosis comment with curl in backticks

ksy36:
Got 406 error when submitting a comment with curl in backticks for a comment like https://bugzilla.mozilla.org/show_bug.cgi?id=1913899#c6:

So stripping just the backticks for curl seem to work. Also added html escape for html elements, like `<iframe>`.

Checklist for reviewer:

### DIFF
--- a/jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
+++ b/jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
@@ -1265,6 +1265,14 @@ class DiagnosisTask(HackbotTask):
                         ]
                         if result.evidence:
                             comment_parts += ["", "Evidence:", "", result.evidence]
+
+                        footer = (
+                            "If you'd like to provide feedback on this comment, please use the 👍 or 👎 "
+                            "reaction."
+                        )
+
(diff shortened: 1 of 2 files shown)


jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -680,6 +681,11 @@ def format_user_story_value(value: str | list[str]) -> str:
     return value
 
 
+def sanitize_bugzilla_comment(text: str) -> str:
+    """Unwrap backtick-quoted curl commands in agent-generated comment text."""
+    return re.sub(r"`(curl\b[^`\n]*)`", r"\1", text)
```
jgraham, line 686: I don't understand why we have this particular special case. Do we know if it corresponds to something specific in the bugzilla code or in the WAF configuration? My concern is that we'll find there's some other cases which are basically the same which we'll run into in the future if we don't understand why this case exists.

jgraham (changes requested):

PR #585, opened 2026-09-03
feat(highwind): write experiment and run level errors to BigQuery

jaredlockhart:
Because

* The job evaluates a grid of cells per experiment and writes every one of them, so a cell that failed already carries its state and its error into BigQuery where it can be queried.
* Failures that happen outside a cell have no row to land in: an experiment that fails before producing any cells, a recipe the run refuses to analyse, and the run-level account of what was consumed and what looked wrong.
* Those reach stdout only. Under Airflow that is an account nobody queries, and the run that dies partway is the one whose reason is worth the most.

This commit

* Replaces the analysis module's prints with a logger, configured once in the entry point so a container's logs read as the run report they did before.
* Adds a third table in the same dataset, written as one load job that replaces the run date's partition like the other two.
* Writes it in a `finally`, so a run that raises still records its own cause.
* Follows the same rule as the other tables about which runs may write: a filtered, sampled, local or validation run writes no partition it could replace a full run's with.
* Takes the log table's columns from the log handler Jetstream already writes through, so adopting that handler later is a swap rather than a rewrite.

`analysis_basis`, `segment` and `analysis_period` come from that handler and are declared here but always null, with the reason in each column description: those concepts are specified for this job but not yet implemented, so the column is the seam. Three columns deviate from the handler, each noted in the schema. The slug column is `experiment_slug`, matching this job's other two tables and its clustering field. The timestamp is a real `TIMESTAMP` rather than a formatted string. `exception` carries the formatted traceback rather than a tuple repr, since the line that raised is the diagnosis.

### DIFF
--- a/jobs/highwind/highwind/output_writing.py
+++ b/jobs/highwind/highwind/output_writing.py
@@ -4,11 +4,18 @@ Two destinations, for two different readers. BigQuery keeps the corpus queryable
 which is what makes cross-experiment context and meta-analysis possible and lets a result be
 inspected without opening a blob. GCS is the seam Experimenter reads, matching where the enrollment
 funnel already writes from this same Airflow.
+
+A third table holds what the run logged. A cell that failed is already queryable, because the
+results grid is written whole and carries a state and an error, but an experiment that failed before
+producing any cell has no row to carry one, and neither has a refused recipe or a run-level event.
+Those reach stdout, which under Airflow is the account nobody reads.
 """
 
 import datetime
 import json
+import logging
 import pathlib
+import traceback
 from dataclasses import dataclass
 
 from google.cloud import bigquery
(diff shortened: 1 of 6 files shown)


jobs/highwind/highwind/output_writing.py
```diff
@@ -233,17 +246,113 @@
     ),
 ]
 
+# One row per log record the run emitted. The columns after the first two are the ones Jetstream's
+# own BigQuery log handler writes, names included, so that adopting that handler here later is a
+# swap rather than a rewrite of everything reading this table.
+LOG_SCHEMA = [
+    bigquery.SchemaField(
+        "as_of_date",
```
mikewilli, line 254: Maybe this?
```suggestion
        "run_date",
```

jobs/highwind/highwind/output_writing.py
```diff
@@ -333,6 +442,92 @@ def write_tables(client, as_of, results_by_slug, cells_by_slug, outputs):
...
+
+def write_log_table(client, as_of, rows, outputs):
+    """Write what the run logged as the whole contents of its date's partition.
+
+    Replaced rather than appended for the reason the tables above are, and with the same
+    consequence: a retry's log replaces the log of the attempt it retried, so the partition
+    describes the run that produced the day's results rather than every attempt at them.
+    """
+    ensure_table(client, outputs.log_table, LOG_SCHEMA)
+    replace_partition(client, outputs.log_table, LOG_SCHEMA, rows, as_of)
```
scholtzan: I don't think we'd want to replace the log table partition. This will overwrite logs from runs that happened on the same day, e.g. a rerun overwrites an earlier runs logs -> we might loose some useful information if the rerun was due to previous errors. I'd treat the log table as append-only.

mikewilli (approved):
Seems fine, and this is another thing that should come basically for free when implementation starts in Jetstream since this is a capability there already.

The "partition" terminology not matching with BigQuery partitioning is throwing me off, so maybe that's something to rename for production (unless I'm misunderstanding).

jobs/highwind/highwind/output_writing.py
mikewilli: Yea good point. Maybe instead of logging with a run date tag, it can be tagged with some other run identifier that allows for multiple attempts on the same date?

mikewilli: (or maybe the run date + timestamp is good enough to differentiate)

jaredlockhart:
@mikewilli yeah exactly, the way I'm starting to think about it is like

* Start reshaping the Highwind job closer to Jetstreams shape
* I haven't started yet but soon I'll start trying to do a similar thing on the Jetstream side to start splitting it apart into more modular pieces to make a landing zone

And if we do both sides of those a bit, then we should be able to fold them together more cleanly, and I think the grounding point there will be when we start migrating all the metrics, ideally we do that in one place and both Highwind and Jetstream should pick them up together.

mikewilli:
Sure, I think it's pretty amenable to another code path right now (with a new click sub-command and a highwind executor), but happy to discuss if you had other thoughts. Also happy to meet if you still want to go through the jetstream codebase together.

PR #589, opened 2026-09-06
Store bugzilla autowebcompat comments ids

ksy36:
Checklist for reviewer:


jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -1257,12 +1293,23 @@ def populate_updates(self, updater: Updater) -> None:
...
+                        footer = (
+                            "If you'd like to provide feedback on this comment, "
+                            "please use the 👍 or 👎 reaction."
                         )
 
+                        comment_parts += ["", "---", "", footer]
+
+                        bug_update.add_comment = bugzilla.CommentCreate(
+                            comment="\n".join(comment_parts)
+                        )
```
ksy36, line 1314: I'll add another PR for storing and polling reactions for hackbot diagnosis comments, but I think this footer can already be added as we won't need a backfill.

jobs/webcompat-kb/data/sql/autowebcompat/tables/autowebcompat_bugzilla_comment/meta.toml
```diff
@@ -0,0 +1,3 @@
+name = "autowebcompat_bugzilla_comment"
```
jgraham: I think just `bugzilla_comments` would be fine given the dataset is already `autowebcompat`.

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -652,11 +673,26 @@ def fetch_data(self) -> None:
...
 
     def update(self) -> None:
-        for _, bug_update in self.bug_updates.values():
-            if bug_update.has_updates():
-                self.client.update_bugs(bug_update.bug)
-            for attachment in bug_update.add_attachments:
-                self.client.create_attachment(attachment)
+        for bug_id, (_, bug_update) in self.bug_updates.items():
+            try:
+                if bug_update.add_comment is not None and bug_update.run_id is not None:
```
jgraham: Can `run_id` actually be `None` in this case? What would it mean to have a comment but no `run_id`?

Also should this go below, at least under the `if bug_updates.has_updates()` block?

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -1257,12 +1293,23 @@ def populate_updates(self, updater: Updater) -> None:
                             "",
                             result.root_cause,
                         ]
+
                         if result.evidence:
                             comment_parts += ["", "Evidence:", "", result.evidence]
-                        bug_update.bug.comment = bugzilla.Comment(
-                            body="\n".join(comment_parts)
+
+                        footer = (
```
jgraham: I mean OK< but Id definitely just have written a multi-line literal here and appended it to `comment_parts` rather than creating yet another intermediary list.

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
```diff
@@ -1357,6 +1404,8 @@ def run(
             if task_runner.has_updates():
                 task_runner.populate_updates(updater)
         updater.update()
+        if isinstance(updater, BugzillaUpdater) and updater.posted_comments:
```
jgraham: I mean, OK, but I might just give each updater a `record_updates` method that takes the bqservice rather than having a special case here.

jgraham (changes requested):

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat.py
ksy36: > Can run_id actually be None in this case? What would it mean to have a comment but no run_id?

Mmm no, a comment always comes from a diagnosis run, so there always be a `run_id`. I only made it optional because `BugUpdate` is constructed empty for every fetched bug, before we know whether it'll get a comment. I've replaced the `run_id is not None` check with an assert. 

> Also should this go below, at least under the if bug_updates.has_updates() block?

So sometimes the comment request fails because of bugzilla's WAF, I added it first before attachment and tags, so it doesn't keep readding them on the next run and fail early. Assuming this won't be a problem soon though, I need to file a bug so it gets fixed on the bugzilla side :)

jgraham (approved):

PR #591, opened 2026-09-08
Autowebcompat feedback etl

ksy36:
Checklist for reviewer:


jobs/webcompat-kb/data/sql/autowebcompat/tables/bugzilla_comments/table.toml
```diff
@@ -8,4 +8,8 @@ mode = "REQUIRED"
 
 [run_id]
 type = "STRING"
+mode = "REQUIRED"
+
+[is_readable]
+type = "BOOLEAN"
 mode = "REQUIRED"
```
ksy36, line 15: Since `is_readable` is REQUIRED, so I need to do a manual migration before this deploys.

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat_feedback.py
```diff
@@ -0,0 +1,231 @@
+"""Collect feedback on Bugzilla comments posted by autowebcompat.
+
+Reactions aren't part of a bug's history and don't update its last_change_time,
+so this job is polling the comments we know about and record their current state.
+"""
```
ksy36, line 5: A separate ETL is probably overkill, but this doesn't seem to belong in either the autowebcompat or bugzilla jobs :)

jobs/webcompat-kb/data/sql/autowebcompat_feedback/meta.toml
```diff
@@ -0,0 +1,2 @@
+name = "autowebcompat_feedback"
```
jgraham: Creating a whole dataset for this feels like overkill. I guess the idea is that the `import_runs` table is unique per dataset. But I think I'd go for having this in the `autowebcompat` dataset and either just having a single ETL that does both things, or a second table like `import_runs_feedback`.

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat_feedback.py
```diff
@@ -0,0 +1,231 @@
...
+
+    Bugzilla fails the whole request if any comment in it was deleted, is
+    private, or belongs to a bug we can't access, and the error only reports
+    the first id it objected to. So ask for them all at once, and only if that
+    fails ask one at a time to find out which ones to give up on."""
+    try:
+        return FetchResult(
+            comments=bz_client.comments_as(list(comment_ids), CommentWithReactions)
+        )
+    except Exception as e:
```
jgraham: This should explicitly be a `BugzillaError`, right?

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat_feedback.py
```diff
@@ -0,0 +1,231 @@
...
+            else:
+                logging.warning(
+                    f"Skipping comment {comment_id} this run: {error_message(e)}"
+                )
+                result.failed_ids.append(comment_id)
+        time.sleep(1)
+    return result
+
+
+def to_rows(
```
jgraham: You can probably just set this up as custom serialisation options on the `comment_with_reactions` type.

jobs/webcompat-kb/data/sql/autowebcompat/tables/bugzilla_comments/table.toml
jgraham: The other option would be to put the `is_readable` field in the `bugzilla_comment_reactions` table, so we create a row for every comment, but only record reactions for ones that are readable. Teh advantage is that the `bugzilla_comments` table doesn't have shared ownership in the sense that two different ETLs are required to fully populate it, and we don't have to pre-populate with fake/assumed data in order to meet the validity requirements. It also means that we don't end up re-writing both tables every time there's an unreadable comment.

The disadvantage would be that consumers would have to know to add `WHERE is_readable` to their queries to only get comments where we have valid reactions data.

Nevertheless I think we should move this into the same table as the reactions data.

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat_feedback.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+class CommentWithReactions(BaseModel):
+    model_config = ConfigDict(extra="ignore")
+
+    id: int
+    reactions: dict[str, int]
+
+
+def error_body(error: Exception) -> Mapping[str, Any]:
```
jgraham: So I'm a bit confused. With bugdantic I don't think we're going to get a `response` object on a bugzilla error at the moment. We're just raising a `BugzillaError` with a `message` property. This is maybe a bug in bugdantic that we're dropping the other error information, but nevertheless. Is this some different kind of exception here? It would help if the code was more specific about what kind of class it's actually expecting rather than doing introspection on a generic Exception type.

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat_feedback.py
```diff
@@ -0,0 +1,231 @@
...
+        return FetchResult(
+            comments=bz_client.comments_as(list(comment_ids), CommentWithReactions)
+        )
+    except Exception as e:
+        logging.info(
+            f"Request for {len(comment_ids)} comments failed ({error_message(e)}), "
+            "retrying them one at a time"
+        )
+
+    return fetch_comments_individually(bz_client, comment_ids)
```
jgraham: So AFAICT we try to fetch all comments ids, and then fall back to getting one at a time if there is any error. But one-at-a-time could end up being thousands of requests to bugzilla.

For old comments we already have some information about whether we think they're readable or not. If that information is usually correct we could split the work in two: re-reading the data for known comments (likely to succeed) and reading the data for new comments (might fail). But that might not be a good assumption; maybe bugs are moved into private states often enough that the cached data might be wrong as often as new bugs are private. In that case maybe we should work with chunks at the application layer, so that we try to fetch e.g. 100 comments at a time and then fall back to individual fetches for that chunk if it fails. 

Another concern is that the cached information about existing comments being unreadable might go out of date e.g. if the bug is opened up. That probably doesn't matter loads for our use case, but does mean that we might be missing some data for e.g. sec bugs.

jgraham (changes requested):

jobs/webcompat-kb/data/sql/autowebcompat_feedback/meta.toml
ksy36: Sounds good, I've changed it to be part of autowebcompat dataset

jobs/webcompat-kb/data/sql/autowebcompat/tables/bugzilla_comments/table.toml
ksy36: So I added a separate table, `bugzilla_comments_unreadable`, which stores unreadable ids and the error code. That way it just appends unreadable ids to it and overwrites `bugzilla_comment_reactions` with WRITE TRUNCATE, without having to preserve is_readable=false records.

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat_feedback.py
ksy36: Good point, I've added chunking. 

> Another concern is that the cached information about existing comments being unreadable might go out of date e.g. if the bug is opened up. That probably doesn't matter loads for our use case, but does mean that we might be missing some data for e.g. sec bugs.

True. I decided to store unreadable ids in `bugzilla_comments_unreadable` with error_code + when the failure was recorded. If we need to we can refetch these comments if some time has passed, i.e. a month or 2 months.

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat_feedback.py
ksy36: I've added the code part and changed generic `Exception:` to `httpx.HTTPStatusError` on bugduntic side in https://github.com/jgraham/bugdantic/pull/60/

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat_feedback.py
ksy36: Yeah, changed it to `BugzillaError`

ksy36:
I've updated bugdantic version, would you be able to take another look please, @jgraham ?

jobs/webcompat-kb/webcompat_kb/etl/autowebcompat_feedback.py
```diff
@@ -0,0 +1,197 @@
...
+    return result
+
+
+def mark_unreadable(
+    bq_client: BigQuery,
+    unreadable_table: TableSchema,
+    unreadable_ids: Mapping[int, int],
+) -> None:
+    """Record that we should stop requesting these comments"""
+    now = datetime.now(tz=UTC)
```
jgraham, line 132: You could have an early return for the perhaps common case of no unreadable ids.

jgraham (approved):

PR #597, opened 2026-09-14
Prioritized_kb_entries: Add total_score for each metric, size estimate and platform teams

benoit-chauvet:
…timate and platform teams

Checklist for reviewer:


jobs/webcompat-kb/data/sql/webcompat_knowledge_base/views/prioritized_kb_entries/view.sql
```diff
@@ -14,6 +14,12 @@ WITH
     site_reports.site_rank_score,
     site_reports.intervention_score,
     site_reports.triage_score,
+    site_reports.user_story,
+    site_reports.score,
```
jgraham, line 18: So these days `score` and `triage_score` are the same thing (and `severity_score` is basically never used). You don't need to do the cleanup here, but maybe we should in general.

jobs/webcompat-kb/data/sql/webcompat_knowledge_base/views/prioritized_kb_entries/view.sql
```diff
@@ -14,6 +14,12 @@ WITH
...
+    {% for metric in metrics.values() if metric.site_reports_condition('host_categories') -%}
```
jgraham: It doesn't actually break anything but it would be easier to follow if we used the same table name argument to `metric.site_report_conditions` for the check as we do when constructing the actual field.

jobs/webcompat-kb/data/sql/webcompat_knowledge_base/views/prioritized_kb_entries/view.sql
```diff
@@ -42,7 +48,23 @@ WITH
...
+      "platform_score", kb_reports.platform_score,
+      "impact_score", kb_reports.impact_score,
+      "affects_score", kb_reports.affects_score,
+      "site_rank_score", kb_reports.site_rank_score,
+      "intervention_score", kb_reports.intervention_score,
+      "triage_score", kb_reports.triage_score,
+      "diagnosis_team", lower(trim(json_value(kb_reports.user_story,'$.diagnosis-team')))
+      )) AS reports,
+      {% for metric_type in metric_types if metric_type.name == 'total_score' -%}
+        {% for metric in metrics.values() -%}
```
jgraham: I would use `{% for metric_name, metric in metric.items() %}` and then use `metric` and `metric_name` in the line below (notably `metrics[metric.name] == metric` so we don't need to look it up again).

jobs/webcompat-kb/data/sql/webcompat_knowledge_base/views/prioritized_kb_entries/view.sql
```diff
@@ -42,7 +48,23 @@ WITH
...
+      "bug", kb_reports.site_report_id,
+      "url", kb_reports.site_report_url,
+      "priority", kb_reports.site_report_priority,
+      "platform_score", kb_reports.platform_score,
+      "impact_score", kb_reports.impact_score,
+      "affects_score", kb_reports.affects_score,
+      "site_rank_score", kb_reports.site_rank_score,
+      "intervention_score", kb_reports.intervention_score,
+      "triage_score", kb_reports.triage_score,
+      "diagnosis_team", lower(trim(json_value(kb_reports.user_story,'$.diagnosis-team')))
```
jgraham: We're not 100% consistent but mostly try to use uppercase for SQL functions.

jgraham (changes requested):
Looks like it probably works fine, just some minor changes.

benoit-chauvet:
@jgraham do you see any problem with adding the metric flags (is_core, is_japan_1000, ...)  in the site_reports JSON payload?
(I'll need this info in my dashboard query, and it will save an extra join)

jgraham:
> @jgraham do you see any problem with adding the metric flags (is_core, is_japan_1000, ...) in the site_reports JSON payload?

Do you mean the `site_issue_reports`  field in this view or the actual `webcompat_knowledge_base.site_reports` view? The former is fine, the latter is not (that view doesn't contain anything that depends on host rankings).

benoit-chauvet:
> > @jgraham do you see any problem with adding the metric flags (is_core, is_japan_1000, ...) in the site_reports JSON payload?
> 
> Do you mean the `site_issue_reports` field in this view or the actual `webcompat_knowledge_base.site_reports` view? The former is fine, the latter is not (that view doesn't contain anything that depends on host rankings).

I mean the  the `site_issue_reports` field in this view indeed. So I'll add it to my next commit

jgraham (approved):

